PR #574, opened 2012-04-02
Add Edge Eraser

purplebleed:
Hi,
I am Leo Lin
I just finish a new feature call Edge Eraser for the coding task of GSoC, it will remove the node when you select source and target.


eduramiba:
Thanks, closing old request though. Edges can be removed in data laboratory.

HijackHornet:
I was looking for this feature... Datalaboratory edge deletion is not as practical as doing it in the overview visually

PR #724, opened 2013-01-19
Fix #723: Catch blank nodes during CSV import

### LINKED ISSUE #723: Catch blank nodes during CSV import
When a CSV file is imported using "Import Edges Table" in the Data Laboratory, empty edges are converted to nodes with a blank IDs. Gephi can handle nodes with blank IDs, but trying to export the graph into an XML-based format will throw an exception. This can be very hard to pin down.

shepdl:


eduramiba:
Thank you Dave,
I merged this manually.

PR #745, opened 2013-02-18
Fix GML import - try to keep column format double, if the parser detected it as double

libor-m:
``GML`` is quite a convinient format when creating graphs from scripts. I need to pass several numerical attributes together with the nodes and edges, and this patch removes the pain of manually clicking ``copy column`` and changing it's type to ``Double`` for each of the columns.

Not much error checking is done, but in script generated graphs typo errors do not happen very often;)


mbastian:
Thanks for the request but there was some recent changes in the API so I had to reproduce the fix elsewhere is #1151. Closing this request then.

Issue #572, opened 2012-03-29
Preview outputs margin automatic fitting


sheymann:
Following https://forum.gephi.org/viewtopic.php?t=1762&p=5996#p5996

Output in PDF/PNG/SVG should fit automatically, at least for the nodes.

jersub:
Related to issue #239

jersub:
Automatic fitting is fixed except for labels, cf. #1218.

PR #771, opened 2013-04-15
addressed issue #572

vijeshm:
If the user provides a margin that sufficiently contains the graph, the graph is exported as is. If the provided margin trims the graph, the approach will progressively increase the margin till it is completely contained. Apart from fitting the nodes, edges are also fit. The solution works if we have the rendered image in the form of a linearized pixels array. Unfortunately, I couldnt find anything equivalent to this structure in the case of PDF and SVG. Hence, I was not able to apply it in those cases.


jersub:
Let's link to issue #572

mbastian:
Closing as addressed in #1209

PR #814, opened 2013-08-27
Special case processing for StringList with Regex columns

wmcelderry-etx:
I've added some code to the 0.8.2 branch to allow the company I work for to manipulate StringLists better and we would like to make it available to the community in the hope it may be useful.

We are currently investigating using Gephi internally and would like to establish how we can contribute any further development back to the community if we take this route - bearing in mind the 0.9 upgrade/overhaul.


eduramiba:
Thanks, closing old issue though as 0.9.1 is the current version.

Issue #147, opened 2011-10-22
graphml `edgedefault` not handled to spec


mbastian:
from http://graphml.graphdrawing.org/primer/graphml-primer.html#GraphGraph

	Graphs in GraphML are mixed, in other words, they can contain directed
	and undirected edges at the same time. If no direction is specified
	when an edge is declared, the default direction is applied to the edge.
	The default direction is declared as the XML Attribute edgedefault of
	the graph element. The two possible value for this XML Attribute are
	directed and undirected. Note that the default direction must be
	specified.

However, when specifying an edgedefault Gephi complains "Edge type doesn't
fit with default".

More importantly having a graph with `edgedefault="undirected"` but all edges
but one with `directed="true"` makes the unspecified edge directed as well. My
work around for now is to just not specify edgedefault for the graph, and manually
specifying `directed` as `true` or `false` for each edge - even though this doesn't
strictly follow the GraphML spec.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2010-11-16T22:02:30Z
 * owner: sebastien.heymann
 * assignee: mathieu.bastian
 * the launchpad url was https://bugs.launchpad.net/bugs/676272

eduramiba:
Fix merged

PR #815, opened 2013-09-03
Import GraphML files correctly (Issue #147)

virtuald:
Looking at ImportContainerImpl and ImporterGraphML (and other importers), it appears that the problem is a problem of terminology. In GraphML, the 'default' type means that edges should be set to X by default, otherwise to whatever the specific setting is. In the importer, it appears to use the default as the *only* edge type, and forbids all other edge types.

Diving into the processor code, AppendProcessor.java checks the default edge type, and creates a graph according to that edge type, which does not appear to support other edge types. This would reflect the viewpoint of the importer. 

If you change the default edge type in the GraphML importer to MIXED, and set the edge type based on the default edge type, then the bug is fixed. Looking at other file formats, this adjustment should be made to other file formats that also support mixed representations.


eduramiba:
Thank you for the fix!

PR #857, opened 2013-11-21
Turn off the verbosity of DefaultProcessor

dustalov:
It's pretty hard to use Gephi Toolkit in scripting when `DefaultProcessor` writes unnecessary and non-demanded information about the processed data. This commit turns off the verbosity and makes a better experience of the embedded Gephi.


eduramiba:
Hi Dmitry, we no longer use System.out here, and now use java Logger (81bff5ab01d8a10f8fef18f07459e5c1ca6b02ff).

We will remove all other uses of System.out.
This will make it possible to disable Gephi-specific logging completely.

Thank you for the pull request.

mbastian:
Yes, and DefaultProcessor specifically was fixed in 81bff5ab01d8a10f8fef18f07459e5c1ca6b02ff

PR #863, opened 2013-12-13
Improved ForceAtlas2

jacomyma:
Hi there. This new version should still be tested, but it is far more fluid and efficient in every version I have tested. The improvement is very visible.

It might be branded "ForceAtlas 2.1" or some different name, though. Only the adaptive speed has been changed, but some results might be different due to convergence issues. It worked very well on all the various graphs I use for tests, but some specific graphs might not converge contrary to ForceAtlas2. I did not deal with this since I was not very efficient at understanding how the plugins are integrated. I just modified the algorithm itself.

Also, because more and more computers deal well with multiple threads, I updated the defaut setting from 2 threads to 4 threads.

### DIFF
--- a/modules/LayoutPlugin/src/main/java/org/gephi/layout/plugin/forceAtlas2/ForceAtlas2.java
+++ b/modules/LayoutPlugin/src/main/java/org/gephi/layout/plugin/forceAtlas2/ForceAtlas2.java
@@ -81,6 +81,7 @@ public class ForceAtlas2 implements Layout {
     private double scalingRatio;
     private double gravity;
     private double speed;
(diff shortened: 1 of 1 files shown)


mbastian:
I suggest a minor tweak to automatically define the default thread count and use the same formula than in OpenOrd: `Math.max(1, Runtime.getRuntime().availableProcessors() - 1)` where `Runtime.getRuntime().availableProcessors()` is the number of CPUs on the computer.

mbastian:
Ported the patch to the master branch (0.9) in #1158, therefore closing this request. Thanks!

Issue #654, opened 2012-07-14
Node label outlines offset in Preview mode


oxling:
The outline around a node label may be shifted to the left or right. This bug also appears when the image is exported to PNG. It appears to be affected by font size, the length of the text and the zoom level. For example, it's most visible with a long label viewed at less than 100% zoom.

I'm using the the latest code in the master branch on OSX 10.7.

![Screenshot here.](https://github.com/oxling/gephi/blob/3c7031e187145c98a905f20fe7c39f87894c59aa/screenshots/label_issue.png)

PR #1153, opened 2015-10-20
Fix label outline in Preview

jersub:
This fixes both issue #654 (node label outline) and edge label outline.

Before:
before

After:
after


mbastian:
Cool! All looking good, merging. Thanks!

PR #1174, opened 2015-10-29
Updated splash screen

jbliesener:
Created new gif logo. If required, I can also provide the original GIMP
.xcf file. Requires font "Yanone Kaffeesatz" to edit.


jbliesener:
I changed the version number color to red in order to call the user's attention that this is a SNAPSHOT version. In the final release, the version number should return to being blue.

mbastian:
Thanks! Could you share the original xcf file as well, please?

PR #1209, opened 2015-12-12
Compute the preview canvas size from the rendered elements

jersub:
Fixes issue #572 

For each item to render on a Preview target, the canvas size is computed. Then, from the union of all of these rectangles, one gets the global Preview canvas size.

With such an implementation, any Preview plugin can declare the canvas size of its own items to render, so they can be integrated in the global Preview canvas size just like the regular items.

### DIFF
--- a/modules/PreviewAPI/src/main/java/org/gephi/preview/PreviewModelImpl.java
+++ b/modules/PreviewAPI/src/main/java/org/gephi/preview/PreviewModelImpl.java
@@ -41,7 +41,6 @@
  */
 package org.gephi.preview;
 
-import java.awt.Rectangle;
 import java.beans.PropertyEditorManager;
 import java.util.*;
 import java.util.Map.Entry;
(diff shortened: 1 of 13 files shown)


modules/PreviewAPI/src/main/java/org/gephi/preview/PreviewModelImpl.java
```diff
@@ -82,8 +80,7 @@ Development and Distribution License("CDDL") (collectively, the
     //Properties
     private PreviewProperties properties;
     //Dimensions
-    private Dimension dimensions;
-    private Point topLeftPosition;
+    private Rectangle _graphicsCanvas;
```
mbastian: We don't use the _ convention for class variables, should be removed.

modules/PreviewAPI/src/main/java/org/gephi/preview/api/Rectangle.java
```diff
@@ -0,0 +1,148 @@
...
+import java.awt.geom.Rectangle2D;
+
+/**
+ * A <code>Rectangle</code> specifies an area in a coordinate space that is
+ * enclosed by the <code>Rectangle</code> object's upper-left point
+ * <code>(x,y)</code> in the coordinate space, its width, and its height.
+ *
+ * @author Jeremy Subtil
+ */
+public class Rectangle implements Shape {
```
mbastian: I think adding a Rectangle class to the API is confusing. Users don't need any new feature besides the position/width so it shouldn't be in the API. The union is used internally by the `PreviewModelImpl` and can be called in a different way than by subclassing Rectangle.

modules/PreviewAPI/src/main/java/org/gephi/preview/PreviewModelImpl.java
jersub: I put it to try to tell developers not to use this variable directly, `getGraphicsCanvas()` has to be used instead as it defines the variable. Any suggestion for handling this?

mbastian: Add a comment instead, this is implementation code (not API) so just use comments

jersub:
Thanks a lot for the review, I'll handle your comments.

modules/PreviewAPI/src/main/java/org/gephi/preview/PreviewModelImpl.java
jersub: Actually it was a mistake to cache the canvas size. It has to be computed each time to get the last info from the graph model.

jersub:
Done. Sorry for the huge diff again, I amended my previous commit too quickly.

Brief recap: replaced `Rectangle` by `CanvasSize`, removed the canvas size cache from `PreviewModelImpl`.

mbastian:
Ok, thank you, I'll take a look at it tomorrow morning. Did you notice any net performance issues with large graphs? Does the canvas calculation also happen for the Preview Panel (Java2D) or is it only for export? Thanks. Just curious before reading the code :)

jersub:
I haven't done any real performance check for now. The canvas calculation does also happen to the Preview Panel.

jersub:
Execution time for `getGraphicsCanvasSize()` on a big graph: < 30 ms

Called on a first Preview refresh, a zoom reset and an export.

mbastian:
Awesome, thanks for checking! I saw that you left the label calculation for later, this is fine but curious whether it can easily be added. Merging nevertheless.

PR #1231, opened 2015-12-17
Removing any redundant modifiers on interfaces or interface components

kirill-vlasov:


kirill-vlasov:
Closing in favor [#1235](https://github.com/gephi/gephi/pull/1235)

PR #1284, opened 2016-01-10
Feature/db ignore unknown columns refactoring

Kornel:
Hi,

 I see you've @mbastian fixed the issue I've stumbled upon in #1272 - when the column is not known, gephi database import threw an NPE.

 I did a different fix - first, it's always good practice to compare the constant to the field, not the field to the constant, then no null check is necessary. E.g. `if (NodeProperty.ID.equals(foo))` instead of `if (foo.equals(NodeProperty.ID))`. 

 That's the change I did first, but then I also refactored the code a bit - I think it's redundant to iterate over the column names for each result set iteration. Why not fetch the id column metadata first, and then use it?

 Thats my suggestion here, please let me know what you think.

Cheers,
 Kornel


Kornel:
Hey,
 @mbastian any thought on this PR?

K

eduramiba:

Kornel:
No problem, glad I can help.

eduramiba:
Merged, thanks!

Issue #1341, opened 2016-02-05
Closeness in GraphDistance incorrect


jonnywray:
Following on from the brief Facebook thread I checked the definition of closeness in the literature against what is calculated in the GraphDistance statistics plugin. I checked both Newman's Networks book and Brandes and Erlebach, Network Analysis.

Closeness is incorrect calculated. When normalization is selected that results in the correct calculation of closeness but I'm guessing that isn't what was intended.

PR #1342, opened 2016-02-05
Fixed issue with closeness calculations in GraphDistance

jonnywray:
Issue #1341.

- Fixed issues with closeness calculation. Now agrees with commonly accepted definition.
- Added harmonic closeness centrality as a new statistic. Similar to closeness but with some nicer properties
- Removed reporting of shortest path count. It wasn't correct as it only considered one of potentially many shortest paths between a node pair.

### DIFF
--- a/modules/StatisticsPlugin/src/main/java/org/gephi/statistics/plugin/GraphDistance.java
+++ b/modules/StatisticsPlugin/src/main/java/org/gephi/statistics/plugin/GraphDistance.java
@@ -68,11 +68,13 @@ import org.openide.util.Lookup;
  * of Mathematical Sociology 25(2):163-177, (2001)
  *
  * @author pjmcswee
+ * @author Jonny Wray
  */
 public class GraphDistance implements Statistics, LongTask {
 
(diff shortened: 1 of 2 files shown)


mbastian:
Thanks @jonnywray for the patch. I really appreciate that you went the extra miles and added tests and the harmonic closeness.

LancelotHolmes:
well,thanks for the implementation of calculating the harmonic closeness centrality,but I have several questions here, according to wikipedia's definition for [harmonic centrality](https://en.wikipedia.org/wiki/Closeness_centrality#cite_note-boldivigna2014-8),or the definition by Boldi and Vigna (2014) in [ "Axioms for Centrality"](http://www.tandfonline.com/doi/abs/10.1080/15427951.2013.865686), the formula to calculate the harmonic centrality for directed graph should be

that we count the target node's harmonic centrality by the source nodes and not divide by the reachable path,but according to Newman's "Networks: An introduction", the harmonic centrality should be divided by n-1, which makes me puzzled.

PR #1350, opened 2016-02-07
German Localization

hbruch:
Added german language action, flag and german bundle files. 

However, I did not find a way how to add the german locale for netbeans plugins, so especially a couple of menus are not translated yet. 
After pulling the properties file I converted them from UTF-8 to iso-8859-1. Is there an easy way to pull the bundle files in iso-8859-1 directly? The properties files don't contain any header information any more, I suppose transifex doesn't support headers.


eduramiba:
Hi Holger,
Thanks for the pull request and great work with your translation!

Soon we will synchronize all languages translations from transifex into Gephi and add German as a new language. I will then mark this request as merged.

About properties files being UTF-8, it seems that iso-8859-1 files are supported by transifex if you configure them as a different type: http://docs.transifex.com/formats/java-properties
Not sure if we can now convert them now though.
Anyway, is there any known issue that could happen if we keep using UTF-8?

akuckartz:
:+1: Danke!

eduramiba:
About translating the menus, build gephi with "deployment" maven profile, so it downloads netbeans platform translations.

PR #1407, opened 2016-03-01
Added per-node opacity property

rweeks:
NodeRenderer now uses the Node's alpha value when rendering the node in SVG, G2D and PDF outputs. It only uses this value if the new per-node opacity property is checked, so existing behaviour is preserved by default.

It's the answer to my SO question [here](http://stackoverflow.com/questions/35731271/can-the-gephi-node-renderer-preserve-the-original-node-opacity).

The justification for this is, I have a large graph from Spark GraphX that I'd like to visualize. I render one visualization for the full graph, and then I want to render separate visualizations for subgraphs. But I want to preserve the positions of the vertices from the original graph, to maintain the visual context.

One way to do this is to set the alpha of the hidden vertices to 0 and then call the OpenORD layout with the same randomSeed as I used for the full graph visualization.

### DIFF
--- a/modules/PreviewAPI/src/main/java/org/gephi/preview/api/PreviewProperty.java
+++ b/modules/PreviewAPI/src/main/java/org/gephi/preview/api/PreviewProperty.java
@@ -101,6 +101,14 @@ public class PreviewProperty {
      * 100 means opaque.
      */
     public static final String NODE_OPACITY = "node.opacity";
+
+    /**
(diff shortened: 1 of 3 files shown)


eduramiba:

mbastian:
LGMT

juliannej:
Sorry for this basic question, but how do I activate the per node opacity?

PR #1435, opened 2016-03-23
Duplicate statement in ExporterDL.java

kilida:
The `if` body is equivalent to the `else` body.


eduramiba:
Thanks for the pull request. Maybe this is a actually a bug, will need to review.

eduramiba:
Thanks, merged!

PR #1487, opened 2016-05-13
Fix - add try/finally logic to wrap graph reentrant locks for layouts

trajar:
Graph layouts use graph read locks without proper try-catch logic as per the jdk documentation.  When the layouts generate a runtime exception the result leaves the graph in a locked state, which then blocks all access threads (including swing/javafx).


eduramiba:

eduramiba:
Merged in https://github.com/gephi/gephi/commit/4915870367db5b4d425591b0eb49663e478544f5, thanks!

PR #1733, opened 2017-06-13
Negative Weight Support Added

astraey:
Modified the ForceAtlas and ForceAtlas2 layouts so it supports negative weights.

The main idea is that if a negative weight is detected, we redefine the minimum edge value to 0, and apply the corresponding operations to maintain the graph’s data integrity.

Bug Reproduction:
[Youtube Link](https://youtu.be/IZNaEpLIPhA)
After Fix:
[Youtube Link](https://youtu.be/xw5BETS5U0o)


eduramiba:
Thanks for the request, we will need to review it.

vlado-s:
Hi,
could you please clarify what you mean by "if a negative weight is detected, we redefine the minimum edge value to 0"? Does it rescale all edge weight such that you add a constant of the absolute value for the most negative edge weight => that one becomes 0 and all others become more positive?
Thanks'

PR #1736, opened 2017-06-15
Add support for Duplicate Workspace and Merge Workspaces

PZerua:
Added two more actions in Workspace tab to allow the duplication of a workspace and the merge of two workspaces. Language showed is Spanish but all languages are supported.

Duplicate example:

Merge example:
merge


eduramiba:
Thanks for the request, will review as soon as possible :)

PR #1739, opened 2017-06-20
ImporterDOT supports # comments

krk:
A line beginning with a '#' character is considered comment in DOT.


eduramiba:
Thanks for the request, merged!

PR #1777, opened 2017-09-01
(refactor) some opportunities to use lambda functions

desmorto:


eduramiba:
Hi, thanks for the request but our code is Java 7 level at the moment. Maybe we will upgrade on a future release

duncdrum:
i think java8 has arrived so *yay* for lambdas

mbastian:
Let's close this one. Lambdas are great but this is tool old to merge, due to conflicts.

Issue #1505, opened 2016-06-01
Preview Screen is not responsive in Gephi 0.9.1


goodgogol:
Hi,

I am pretty new to gephi and using it to generate network graphs. While the graphs are being generated properly and can be seen in the overview tab, the preview screen remains empty and just shows a black screen. The export options are also failing ,possibly due to this(pdf /svg /Loxa Web Site export all shows a black screen and no graph).
Also i can't spot the partition and ranking options, just the layout and appearance panels are available.Maybe I'm missing something?
Would really appreciate any help in resolving these issues.

eduramiba:
Hi, partition and ranking are now appearance.
About preview, is there any exception in your messages.log file?

Check https://github.com/gephi/gephi/wiki/Troubleshooting to find user directory

goodgogol:
Hi,
Thanks a lot! Downloaded and reinstalled 0.91 and now the previews are working fine and so is the export.
However, still can't find partition and ranking options!

eduramiba:
Hi,
Well appeareance is mostly the same, but the UI has changed slightly (check the icons at the top-right corner and specially the bottom-left icon which allows to switch between ranking and partition).

Also please note that there is a known bug: #1459

goodgogol:
Thanks a ton !Got the ranking ans size options now and they are working fine.
Can you please let me know if this is the case with the OS X release as well?

eduramiba:
Yeah it's the same

darcyof:
Hi, I am also having a problem with the Preview screen.  I am able to make graphs in the overview but am getting only a grey screen on the preview tab.  There is not refresh button (or any other button).

any help will be much appreciated.

Thanks!

muthulakshmi1990:
Hi, I am using Gephi0.9.1. I am trying to connect Titandb with gephi. I am able to see the data from Titandb to Gephi(using DataLaboratary tab in gephi). But I am not able to view graph in Overview or Preview tab in Gephi. Kindly help me to resolve my problem.

saturnjiang:
Hi I am using Gephi0.9.1. when I import the csv data into the datalaboratary table, but I cannot see the graph in the Preview. The screen is black, after I zoom in or out, the graph can be shown shortly with nodes only, then the screen returnd to black. But when I import the data of examples, the preview is normal.
![black screen](https://cloud.githubusercontent.com/assets/25746920/22889965/285691dc-f20b-11e6-9608-72ebe7b9f19f.png)
[...]

PR #1823, opened 2017-10-19
Preview Panel keeps refreshing

YikeShen:
Hello,
I tried to view my figure in Preview panel. 
The preview panel keeps refreshing even after I closed Gephi and change the Memory to 1400m/1024m according to issue #1505

I uninstalled and reinstalled; restarted my MacBook Pro and it's still refreshing. I was getting preview two days before, it's just stopped working today and yesterday. 
Any help will be greatly appreciated.
Thank you!
Sincerely,
Yike

Follow up, I got a version down and the problem is solved. Preview is working now. But there are some problems with showing the labels of the nodes. I doubled checked I had nodes label in my nodes input datasheet. It's fine, I don't have that much nodes. However, if the frozen refreshing can be fixed in the new version, that could be great.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -19,7 +19,8 @@
 
 ## Your Environment
 
-* Version used: Gephi 0.9.1
-* Operating System:
+* Version used: Gephi 0.9.2
+* Java version: 
+* Operating System: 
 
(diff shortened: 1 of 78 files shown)


eduramiba:
Hi,
Should this be an issue instead of a pull request?

YikeShen:
Yes, probably. Thanks@eduramiba

PR #1970, opened 2018-05-31
Graph added to RankingImp upper class and removed from other, update …

### LINKED ISSUE #1968: Node size ranking just on filtered nodes
## Expected Behavior

When you change the size of the nodes depending on some ranking attribute, it should only consider the visible ( =filtered ) nodes. Same if you filter the graph after a ranking sizeing have been made, the sizes must be recomputed to reflect the new sizing depending only on the filtered nodes.

## Current Behavior

When you do a sizeing based on some ranking, it calculates the size of all the nodes in the graph, not only the filtered (= visible) ones.

## Steps to Reproduce

1.
2.
3.
4.

## Context

When there are nodes with high ranking differences and you set the size based on that raking, it is difficult to see the smallest ones. So maybe the biggest ones are not important and you want to make a filter. The result is that the size of the filtered nodes is really similar, because of the high differences with the bigger ones (which are no more in the view, the have been filtered).

## Your Environment

* Version used: Gephi 0.9.2
* Java version: 1.8
* Operating System: Windows 10

Team: @tzadro @guinii

jorgealemangonzalez:
…graph method created and used before callin the functions in appearance controller

fixes #1968 

We think there is an error in all the RankingImpl subclasses, i fixed that error for the DegreeRankingImpl class but not for everyone (maybe there is no error and I didn't undestood the code).

Please give me some feedback if you don't like my implementation.

Thanks.

### DIFF
--- a/modules/AppearanceAPI/src/main/java/org/gephi/appearance/AppearanceControllerImpl.java
+++ b/modules/AppearanceAPI/src/main/java/org/gephi/appearance/AppearanceControllerImpl.java
@@ -120,6 +120,7 @@ public class AppearanceControllerImpl implements AppearanceController {
                 iterable = graph.getEdges();
             }
             try {
+                function.updateRankingGraphAndRefresh(graph);
                 for (Element element : iterable) {
                     function.transform(element, graph);
                 }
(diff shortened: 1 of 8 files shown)


eduramiba:
Thanks for the pull request! We will review it

jorgealemangonzalez:
Should I close this PR ?

mbastian:
Closing as this is no longer relevant, see details in #1968

PR #1979, opened 2018-06-15
Exporting Statistics reports charts as CSV files

thechristmas:
Gephi's Statistics panel allows users to run a series of algorithms over the loaded datasets with the aim of obtaining a descriptive report offering information about said data.

At the moment, these reports can only be saved in HTML format. 

This pull request presents a proposal for implementing an option to export the reports' generated charts to CSV, so that they can be loaded, manipulated and further developed in other softwares such as Excel.

For follow-up on the matter, please feel free to ask for corrections, doubts or any kind of posible inquiry.


eduramiba:
Thanks for the pull request, this is something we already wanted to have, I will review it as soon as possible.

PR #1980, opened 2018-06-15
Export chart statistics in csv

marcperez06:
Hello, We implement the export of statistics.
We upload a zip with the files we modify.
We develop the changes in a new branch called "newMaster", you can found all the code here: https://github.com/marcperez06/gephi

Thanks.

Collaborators: @marcrabat, @jaumemito
[Files_modified.zip](https://github.com/gephi/gephi/files/2106288/Files_modified.zip)


eduramiba:
Hi, is this related to pull request #1979 ?

thechristmas:
@eduramiba No, we’re working independently, but it seems that the intention behind the feature is the same.

eduramiba:
I see, I guess #1979 should be preferred as it's actually a pull request and this is somehow a zip with changes, but I will review both as soon as possible.

Issue #1935, opened 2018-03-16
Cannot export dynamic attributes


schignel:
## Expected Behavior
When computing dynamic attributes, one would expect to be able to export the values into a spreadsheet to be able to work with the data outside of Gephi.

## Current Behavior
An html graph is produced, but there is no way to access the underlying values.

## Possible Solution
Create a "download data" button that exports the underlying values of the html graph as a spreadsheet.

## Steps to Reproduce

1. Run a dynamic attribute
2. View the results

## Context
This severely limits the utility of the dynamic statistics, since it is impossible work with the computed values or make high-resolution charts, which is required for most publications. 

## Your Environment

* Version used: Gephi 0.9.2
* Java version: jre1.8.0_151
* Operating System: Windows 7

marcperez06:
Hello everybody, mi collegues @jaumemito @marcrabat and I, we found interesting this issue, and want to develop a new feature in statistics pluguin, that let you to export the data in CSV format.

We insert a button near to "save" button in the result page.

Thanks.

eduramiba:
@marcperez06 Great! The code is here https://github.com/gephi/gephi/tree/master/modules/StatisticsPlugin/src/main/java/org/gephi/statistics/plugin/dynamic

ijoel92:
Was this implemented?

Pepe-Marquez:
Is there really still not a solution for this?

PR #1983, opened 2018-06-26
New master

marcrabat:
issue #1980 
We managed to do the pull request correctly since we have sent you the code as a zip file.

In order to minimize the changes in the code, we have done the following:
- Implementation of a static class CSVCreator inside StatisticPlugin that generates the csv data receiving the object XYSeries as a parameter.
- For each class implementing getReport() we have added the csv string that appends the csv data created with CSVCreator after the </HTML> tag, which we have thinked it was the more consistent separator mark.
- Finally we have added in UIComponents the export button as well as its icon which retrieves the the data that comes SIMPLEHTMLReport in order to save the file in the system.

Sorry for the inconvenients.


eduramiba:
Thanks, closing #1980 then and we will review this one.

Pepe-Marquez:
#1935 I guess this is related to this PR.

PR #1995, opened 2018-07-12
Update Gephi plugin URL in README

joelhandwell:


eduramiba:
Thanks for the request!

Issue #2088, opened 2019-03-13
Not all selected nodes/edges are moved/copied to new workspace


jbliesener:
## Expected Behavior
When selecting and copying/moving nodes to other workspaces, all selected nodes and edges should be copied, independently from zoom level and pan location.

## Current Behavior
Not all selected nodes are copied, only those in vincinity of the currently viewed area. 

This occurs, because the iterators in OcTree only iterate over visible octants. If the user changed zoom level or pan location between the node/edge selection and the copy operation, some of the selected nodes can be in invisible octants and therefore will not be returned on engine.getSelectedNodes, leading to them not being copied/moved.

This happens as well if the nodes are selected from a plugin with VizualizationController.selectNodes(), which is my specific use case. However, the root problem is that OcTree.OctantIterator does not iterate over invisible octants. 

## Possible Solution
Use different iterator in OcTree to get selected nodes.

## Steps to Reproduce

See attached video

## Context

My plugin selects nodes in the graph with VizualizationController.selectNodes(). Some of these nodes are on screen, others are not. Trying to copy all selected nodes to a new workspace results in not all nodes being copied.

Tracing down the problem, I disovered that although all NodeModels are correctly marked as selected when SelectionManager.selectNodes() calls engine.selectModels (https://github.com/gephi/gephi/blob/c1d40ce28b482a42e08177f2f6f998da59967cf4/modules/VisualizationImpl/src/main/java/org/gephi/visualization/api/selection/SelectionManager.java#L167), calling engine.getSelectedNodes() immediately afterwards can return a different number. Tracking it down further reveals that OcTree.OctantIterator does not return nodes in invisible octants.

## Your Environment

* Version used: Gephi 0.9.2
* Java version: 1.8
* Operating System:  Windows 10
[Video copy node issue.zip](https://github.com/gephi/gephi/files/2963040/Video.copy.node.issue.zip)

jbliesener:
Any comment on this?

mbastian:
Thank you, good catch! I've merged your PR. It will be included in the upcoming 0.9.3 release. Cheers,

PR #2089, opened 2019-03-13
Fix for #2088

jbliesener:
This should fix issue #2088. Please check if the call to getVisibleOctantsIterator() in updateLOD() should be changed as well:

https://github.com/gephi/gephi/blob/56b8733f03a3aaaa4bcc3ca63744c803f9d7a4c3/modules/VisualizationImpl/src/main/java/org/gephi/visualization/opengl/CompatibilityEngine.java#L451

### DIFF
--- a/modules/VisualizationImpl/src/main/java/org/gephi/visualization/api/selection/SelectionManager.java
+++ b/modules/VisualizationImpl/src/main/java/org/gephi/visualization/api/selection/SelectionManager.java
@@ -131,6 +131,7 @@ public class SelectionManager implements VizArchitecture {
     }
 
     public void setCustomSelection() {
+        engine.setRectangleSelection(false);
         vizConfig.setSelectionEnable(false);
         vizConfig.setDraggingEnable(false);
         vizConfig.setCustomSelection(true);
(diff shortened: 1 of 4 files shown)


eduramiba:
Hi,
Thanks for the pull request.

I will check and merge it as soon as possible. But this code will be replaced with the new [visualization engine](https://github.com/gephi/viz-engine) (work in progress).

jbliesener:
Thank you, Eduardo. This issue has become critical and the change is relatively simple. I understand that it's not a priority for users that only select nodes visually, but as soon as you start using VisualizationController.selectNodes and some of these nodes are out of the currently visible network, you run into issues.

I'd appreciate if you could provide a quick fix through the regular 0.9.2 update channels. Thank you very much.

jbliesener:
Oh, and please ignore the reverted "Install gephi.conf with user write privilege" fix. It has nothing to do with this issue. It was a try to give a non-admin user on Windows write privilege on the gephi.conf file, but I wasn't able to test it due to missing pre-requisites, so I abandoned it. However, I had already commited it to my private repo, so I decided to undo it.

Having said that, if you find some time, you may want to test it...

jbliesener:
New VizEngine looks exciting!

eduramiba:
Sorry for the delay, I need to find the time to merge and release updates for this.

PR #2115, opened 2019-06-26
Macro engine

victorvila92:
As a **students of an Open Source projects course from University Pompeu Fabra (Barcelona)**, we have implemented a first version of a **system that allows to record, edit and execute macros**. We define a macro as a set of user actions that are stored to be easily executed at any time with a single click.

We've created a new menu in tool section that allows a user to:

- See the list of the stored macros
- Edit a macro's name
- Delete a macro
- Execute a single macro

The engine is currently working for actions in Appearance, Layout and Filters modules.


jbliesener:
+1, I like the idea of tracing filters, layouts and appearance actions. This macro engine is a step towards a full execution history, which would allow undos as well.

victorvila92:
> +1, I like the idea of tracing filters, layouts and appearance actions. This macro engine is a step towards a full execution history, which would allow undos as well.

Thx @jbliesener , we appreaciate your feedback

eduramiba:
Thanks for the request, we will evaluate it.

Issue #2340, opened 2021-10-25
Update crash reporter and make it GDPR compliant


mbastian:
The Gephi crash reporter sends crashes to our [Sentry account](https://sentry.io/organizations/gephi/issues/?project=85815). This needs some maintenance as we should be using the sentry-java library and also make sures there are no PII in the report (like username).

### Definition of done
* Crash reporter uses the recommended Sentry library
* Crash reports are GDPR compliant

wetneb:
I am having a go at the update (first bullet point).

PR #2365, opened 2021-12-01
Migrate Sentry to new SDK

wetneb:
For #2340.

This only updates to the new SDK, without solving the problem of logging personal information from users.
It looks like we are already trying to log some personal fields (user name, user email), so a first step would probably be to exclude those? I think @mbastian also had some heuristics to hide user names from paths in mind.

### DIFF
--- a/modules/DesktopBranding/pom.xml
+++ b/modules/DesktopBranding/pom.xml
@@ -100,6 +100,11 @@
             <groupId>io.sentry</groupId>
             <artifactId>sentry</artifactId>
         </dependency>
+	<dependency> <!-- required by Sentry -->
+	    <groupId>com.google.code.gson</groupId>
+	    <artifactId>gson</artifactId>
+        </dependency>
+
         
     </dependencies>
 
--- a/modules/DesktopBranding/pom.xml
+++ b/modules/DesktopBranding/pom.xml
@@ -99,12 +99,6 @@
         <dependency>
             <groupId>io.sentry</groupId>
             <artifactId>sentry</artifactId>
-            <!-- exclusions>
-                <exclusion>
-                    <groupId>org.slf4j</groupId>
-                    <artifactId>slf4j-api</artifactId>
-                </exclusion>
-            </exclusions -->
         </dependency>
         
     </dependencies>
--- a/modules/DesktopBranding/pom.xml
+++ b/modules/DesktopBranding/pom.xml
@@ -97,25 +97,16 @@
             <artifactId>org-openide-awt</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.getsentry.raven</groupId>
-            <artifactId>raven</artifactId>
-            <exclusions>
+            <groupId>io.sentry</groupId>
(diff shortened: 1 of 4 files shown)


wetneb:
Also, it seems that the format of the PUSH_URL has changed in this new API: it is now expecting something of the form `"https://examplePublicKey@o0.ingest.sentry.io/0"`, and we currently have something with `app.getsentry.com` instead of `ingest.sentry.io`. Not having access to the Sentry account I could not update this URL myself. I also wonder if this URL should be made public in the code (as it currently is) or if it should rather be injected at packaging time (as a secret).

wetneb:
This does not work yet: somehow the `com.google.code.gson` library is not reachable at runtime:
```
[INFO] java.lang.NoClassDefFoundError: com/google/gson/GsonBuilder
[INFO] 	at io.sentry.EnvelopeReader.<init>(EnvelopeReader.java:22)
[INFO] 	at io.sentry.SentryOptions.<init>(SentryOptions.java:90)
[INFO] 	at io.sentry.SentryOptions.empty(SentryOptions.java:1560)
[INFO] 	at io.sentry.NoOpHub.<init>(NoOpHub.java:15)
[INFO] 	at io.sentry.NoOpHub.<clinit>(NoOpHub.java:13)
[INFO] 	at io.sentry.Sentry.<clinit>(Sentry.java:24)
[INFO] 	at org.gephi.branding.desktop.reporter.ReportController.<init>(ReportController.java:88)
```

wetneb:
@mbastian this seems good to go for me know. I only tested this manually (triggering a fake exception) and it did get reported in Sentry.

I restored the user email field because it is actually provided on a voluntary basis by the user. If we do not want that, we should remove that field from the UI.

The paths that appear in the exception report do still contain the user name (at least for me), so that is not anonymised, but I do not see an obvious way to do that.

mbastian:
> @mbastian this seems good to go for me know. I only tested this manually (triggering a fake exception) and it did get reported in Sentry.
> 

Awesome! I checked out the exception and everything seems to be nicely organised there.

> I restored the user email field because it is actually provided on a voluntary basis by the user. If we do not want that, we should remove that field from the UI.

OK, sounds good. Let's leave the UI field for now as it's additional work.

> 
> The paths that appear in the exception report do still contain the user name (at least for me), so that is not anonymised, but I do not see an obvious way to do that.

Thanks for taking a closer look. I had found this [capability](https://docs.sentry.io/platforms/java/data-management/sensitive-data/) but it doesn't seem it will do anything for us here. It means we should scrub this PII information ourselves from the logs. My recommendation is to try to cover the most common cases with a simple search/replace based on regex:
- /home/XXX
- \Users\XXX
- /Users/XXX

What do you think?

mbastian (approved):
Looks great

wetneb (commented):
I added some regex to anonymize stuff, with some tests to demonstrate. No guarantee it is bullet proof of course…

mbastian (approved):
Nice work!

Issue #2139, opened 2019-09-10
Node betweenness is negative for big graphs when normalized


wackyworm:
For big graphs, more than 50K nodes, the betweenness value is negative and incorrect when normalized because of integer overflow in this line:

https://github.com/gephi/gephi/blob/56b8733f03a3aaaa4bcc3ca63744c803f9d7a4c3/modules/StatisticsPlugin/src/main/java/org/gephi/statistics/plugin/GraphDistance.java#L365

(n - 1) * (n - 2) is bigger than integer max value for n > 50K, producing negative numbers

eduramiba:
Thanks for your report, adding to 0.9.3 milestone

PR #2366, opened 2021-12-01
Make sure if node count is large,, the betweeness normalization factor isn't negative

totetmatt:
Fix for #2139


mbastian:
Awesome well done! 👍

mbastian (approved):

PR #2399, opened 2022-01-05
Export org.gephi.ui.propertyeditor as public package for plugins

illarionov:


mbastian:
Thanks, I merged master into 0.9.3 so that the snapshot builds are refreshed. If you pull the latest 0.9.3 snapshopts from maven central it should include your change. You nedd to have that repository though:

```
    <repositories>
        <repository>
            <id>oss-sonatype</id>
            <name>oss-sonatype</name>
            <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
            <snapshots>
                <enabled>true</enabled>
            </snapshots>
        </repository>
    </repositories>
```

PR #2425, opened 2022-02-08
Use dynamic URLs for JRE download.

### LINKED ISSUE #2424: Improve JRE download URLs to get the latest release for a given Java version
The packaging process currently relies on static URLs to download JREs for all three platforms.
This means that the bundled JREs might not be as fresh as they could be.
Using recent JREs has various benefits: more up to date root SSL certs, latest vulnerability fixes, and probably some bug fixes from time to time.

We could instead use dynamic URLs which redirect to the latest release of the JRE for a given Java version.
There is an API documented here: https://api.adoptium.net/q/swagger-ui/#/Binary/getBinary

For instance this will return the latest Java 11 JRE for Mac:
https://api.adoptium.net/v3/binary/latest/11/ga/mac/x64/jre/hotspot/normal/eclipse?project=jdk

That would be a good candidate to replace the static link we currently have:
https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.13%2B8/OpenJDK11U-jre_x64_mac_hotspot_11.0.13_8.tar.gz

wetneb:
Closes #2424.

And now I am off to replicating Gephi's amazing setup to OpenRefine!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -167,6 +167,7 @@
         <gephi.update.center.thirdparty.url>https://raw.githubusercontent.com/gephi/gephi-plugins/gh-pages/plugins/${project.version}/updates.xml</gephi.update.center.thirdparty.url>
 
         <!-- JRE locations -->
+        <!-- for more details about the parameters supported by this API, see https://api.adoptium.net/q/swagger-ui/#/Binary/getBinary -->
         <gephi.bundle.jre.macos>https://api.adoptium.net/v3/binary/latest/${gephi.javac.release}/ga/mac/x64/jre/hotspot/normal/eclipse?project=jdk</gephi.bundle.jre.macos>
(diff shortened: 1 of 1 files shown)


eduramiba:
Thanks @wetneb Looks good to me!

eduramiba (approved):

wetneb:
Note: I have a doubt about whether the setup (with or without this PR) works for Windows. Because for Windows, the JRE is downloaded as a `.zip` file, but is then extracted using `tar -zxf`, which should not work. It might be worth checking.

wetneb:
Actually, no it should work when run on Windows, because somehow recent versions of Windows include `bsdtar` as `tar`. Mind blown!

mbastian:
Amazing, I was wondering how to do that. Thanks!

PR #2598, opened 2022-08-31
Make edge kind editable#2518

### LINKED ISSUE #2518: Make edge kind editable
We do support multi-graphs but it's currently not possible to edit this column via the Data Laboratory.

### Definition of done
* Edge kind is properly shown in the 'Edit' panel (under properties)
* Edge kind can be changed from the Data Laboratory 
* Data Laboratory manipulators allow 'kind' to be edited (fill column with value, clear etc.)
* Does not show that column if the graph is not a multi-graph
* New column with `Kind` name can't be added if already present
* (Optional) Creating a `Kind` column as an option in the `Merge Column` menu
* When exporting via CSV, the 'Kind' column should be present when relevant

Notes
- In the [Graph API](https://www.javadoc.io/doc/org.gephi/graphstore/latest/index.html), edge kind is named edge label.
- The edge kind type is defined in the graph configuration and can't be changed

totetmatt:
https://github.com/gephi/gephi/issues/2518

Edge kind is now editable from Datatable. 

How to test : 

Open Gephi with new Workspace
* Create 2 nodes
* Create a link with a type name "A"
* Select the 'Kind' on the Edges Table
* Change "A" to "Z"
* Notice the change is keep on table
* Check on overview , Appearance > Edges > Partition > Type   that now "Z" appears


totetmatt:
Scope has been reduced as Edge kind is quite "special" at the moment and needs a deeper refactor from  graphstore.

modules/DesktopDataLaboratory/src/main/java/org/gephi/desktop/datalab/tables/EdgesDataTable.java
```diff
@@ -92,8 +92,19 @@ public Class getColumnClass() {
 
             @Override
             public Object getValueFor(Edge edge) {
-                return edge.getTypeLabel() != null ? edge.getTypeLabel().toString() : null;
+                return edge.getTypeLabel() != null ? edge.getTypeLabel().toString() : "";
```
eduramiba, line 95: Is this necessary? Let's test the effect of editing the row and setting the empty string again. Does it change the edge type to a new one?

Maybe in `setValueFor` we should check for empty string and in that case `setType(null)`. And even safer, check for empty string after trimming, to not generate weird blank string edge types on accident.

Apart from this, LGTM! 👍🏼

totetmatt: Yeap right, I've changed the behaviour to have empty string that mapped to kind type 0

eduramiba: Thanks @totetmatt!

modules/DesktopDataLaboratory/src/main/java/org/gephi/desktop/datalab/tables/EdgesDataTable.java
```diff
@@ -94,6 +94,24 @@ public Class getColumnClass() {
             public Object getValueFor(Edge edge) {
                 return edge.getTypeLabel() != null ? edge.getTypeLabel().toString() : null;
             }
+
+            @Override
+            public void setValueFor(Edge element, Object value) {
+                String strValue = (String) value;
+                if(!strValue.isEmpty()) {
```
eduramiba, line 101:
```suggestion
                if (strValue != null && !strValue.trim().isEmpty()) {

```

modules/DataLaboratoryPlugin/src/main/resources/org/gephi/datalab/plugin/manipulators/columns/merge/Bundle.properties
```diff
@@ -4,7 +4,8 @@ JoinNumberColumns.name=Join numerical columns
 JoinNumberColumns.description=Join the values of number/number list columns into a new column with the type double[].
 CreateTimeInterval.name=Create time interval
 CreateTimeInterval.description=Create a time interval for each row taking 1 or 2 columns as start/end times.
-
+GenerateAnEdgeKind.name=Use the column to generate a typed edge
```
eduramiba, line 7: Maybe we should leave the merge strategy out of this PR until it's ready, because it will create already resources for translators that are not a real feature yet, and that could be confusing.

modules/DesktopDataLaboratory/src/main/java/org/gephi/desktop/datalab/tables/EdgesDataTable.java
```diff
@@ -94,6 +94,26 @@ public Class getColumnClass() {
...
+                    element.setType(edgeType);
+                } else {
+                    // Type 0 is a special type for "no type defined"
+                    element.setType(0);
+                }
+
+            }
+
+            @Override
+            public boolean isEditable(){return true;}
```
eduramiba, line 114: nit: Formatting ;)

modules/DataLaboratoryPlugin/src/main/resources/org/gephi/datalab/plugin/manipulators/columns/merge/Bundle.properties
totetmatt: Ok fair enough I removed it

eduramiba (approved):
LGTM, thank you!

PR #2638, opened 2022-09-24
#2613 Gather all icons in a central module

### LINKED ISSUE #2613: Better icon management
Related to #2610. We should find a better way to handle icons behaviour because they are currently scattered across several different modules and folder, making them very hard to maintain.

### Proposed solution

Create a module that collects all icons, technical requirements, and control behaviors.

E.g., it can:
- Export individual icons to be imported into other modules;
- Provide variations of icons according to light/dark themes;
- Load themes are designed for visual impairment, children, or, more generally, for better diversity (i.e. RTL).

### Alternatives considered

Just replacing `.png` with `.svg` as attempted during the GephiWeek2022 is a first step but would not improve maintenance or give possibility to switch themes.

### Additional context

We could build on top of an existing icon library to optimise maintenance efforts. We could add icons related to network graphs if needed (pull request).

mbastian:
## Description

In order to streamline the update of icons, we create a unique `DesktopIcons` module with all the icons in it. Icons are organised by module name to easily find them. All UI modules that need icons depend on that module (via standard `pom.xml`)

We also replaced icon loading code with the more modern Netbeans utility `Utilities.loadImageIcon`. This utility supports dark icons automatically and SVGs.

## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [X] 👍 README.md


mbastian:
@eduramiba FYI

PR #2644, opened 2022-09-25
#1388 Warning of newline characters, leading and trailing spaces on elements id

### LINKED ISSUE #1388: Add warning when element IDs match without special characters such as new lines
Hi. I found that 'JohnKomov' and 'JohnKomov\n' are the different names for merging code and the same ones for Gephi itself. The result of such duality is a Java Exception and two nodes with the same visible names and different sets of edges. Check out other non-alphabetical symbols too (I didn't check).

WaltonG:
## Description

Check if elements id has newline characters, leading or trailing spaces and add a warning on the report


mbastian (changes requested):
Hi @WaltonG and thanks for the PR. Some comments:

- You missed adding the change to the `Bundle.properties` file. There needs to be an entry for the `Special_Characters_In_Id` key.
- You only validate the edge list but I think it should be both nodes and edges. I suggest looking at the `checkColorAlpha` function right above. You could make your function generic by using the `ElementDraftImpl` class, which both `NodeDraft` and `EdgeDraft` inherit. The message should however be specific enough so the user knows if it's a node or an edge that is affected.
- It would be great to have unit test for this feature.

Let me know if anything is unclear

mbastian:
Hey @WaltonG do you need any help on this one? Let me know!

WaltonG:
Hi @mbastian, I had not found time to work on this. I'm intending to have
it ready for review by tomorrow.

WaltonG:
Hi @mbastian I have re looked the code but cant seem to get why the test fails

mbastian:
> Hi @mbastian I have re looked the code but cant seem to get why the test fails

Thanks @WaltonG I had a look and made a commit to correct the test. I think the test failed because you were using an ad-hoc `Report` as opposed to the report coming from the container. We actually had a neat function in another class to test the existence of certain issues in a report. I extracted this function out and put it into a `Utils` so we could also use it in this test.

mbastian (approved):
Good to go, thank you @WaltonG!

WaltonG:
Thank you. I'll pick another good first issue

Issue #2342, opened 2021-10-25
Bug bash


mbastian:
As part of the retreat I've selected a bunch of bugs we would like to fix for the next release. We can still update that list of course but good to have something to get started.

#### Import
Typically quite easy to fix when sample files are provided. Good to get started.
* https://github.com/gephi/gephi/issues/2213
* https://github.com/gephi/gephi/issues/2158
* https://github.com/gephi/gephi/issues/2028
* https://github.com/gephi/gephi/issues/1945
* https://github.com/gephi/gephi/issues/1909

#### Stats
Not easy to fix as it usually requires to deep-dive into the reference, or the paper. But quite important as scientists count on accurate results...
* https://github.com/gephi/gephi/issues/2139
* https://github.com/gephi/gephi/issues/2145
* https://github.com/gephi/gephi/issues/1872

#### IO
Typically corrupted .gephi means some bug in the serialization code. This can either be in GraphStore, which serialises the graph or in any other module that provide content for the project file.
* https://github.com/gephi/gephi/issues/2301
* https://github.com/gephi/gephi/issues/2297
* https://github.com/gephi/gephi/issues/2263
* https://github.com/gephi/gephi/issues/2219
* https://github.com/gephi/gephi/issues/2023

### Layout
* https://github.com/gephi/gephi/issues/1816
* https://github.com/gephi/gephi/issues/2202
* https://github.com/gephi/gephi/issues/1714
* https://github.com/gephi/gephi/issues/1829

#### Others
* https://github.com/gephi/gephi/issues/2112
* https://github.com/gephi/gephi/issues/2096
* https://github.com/gephi/gephi/issues/2045

mbastian:
Closing as this bug bash was purely dedicated to the coding retreat.

Issue #1181, opened 2015-11-03
Null pointer Exception while loading data from SQLIte


sumesh1988:
![image](https://cloud.githubusercontent.com/assets/2969687/10901878/f5b24358-821b-11e5-91fc-8729159e0438.png)

I tried loading a table from SQLite.

Test connection is successfull

![image](https://cloud.githubusercontent.com/assets/2969687/10901895/3b8c81b8-821c-11e5-9711-07d9db08c920.png)

Node query: 
select src_email as id from peopleList;

Edge query:
select mail_members.src_email as source, mail_members.trgt_email as target,  count(mail_members.message_id) as weight from mail_members  where mail_members.user_type = 'From' or mail_members.user_type = 'To' group by mail_members.src_email, mail_members.trgt_email  order by weight;

The queries give result in sqlite.

eduramiba:
Does it work with latest 0.9.2 snapshot?

PR #2658, opened 2022-10-07
#1181 Add inverted edge weight feature to ForceAtlas2

### LINKED ISSUE #1816: Edge weigh interpretation in layout algorithms (e.g. ForceAtlas2,...)
## Current Behavior
Currently the interpretation of the edge weight (W) in Gephi (or at least in the ForceAtlas2 layout algorithm) is such, that higher W is interpreted as "stronger connection" and is reflected also in the thickness of the line representing the edge. However other algorithms often interpret W in a inverse fashion, where W is the cost/penalty for "going through" the edge, e.g. when searching for shortest paths (notably the NetworkX routines). Neither way is right or wrong - it is more a "philosophical" question which interpretation will be canonical. But this discrepancy results in some counter-intuitive consequences, such as e.g. when I obtain the shortest path from NetworkX and look at it in Gephi (after node distribution with ForceAtlas2), then it passes through distant nodes connected by thin lines.

It would be therefore beneficial to include an option that would make it possible to obtain a graph layout with the inverse interpretation of W.

## Expected Behavior
The expected behaviour would be such, that nodes connected via edges with smaller W would have stronger attraction than those with larger W. Also the line thickness should be inverted to current state.

## Possible Solution
I have discussed this (via e-mail) with Mathieu Jacomy, who confirmed the above mentioned interpretation. He suggested that I may create a pull. However, I’m afraid I’m not confident enough to modify someone’s else’s code and resume responsibility for it working correctly. There are however two simple solutions for addressing this issue that come to mind:

1. Use the inverse of W in the calculation of the attractive force between connected nodes (eq. 6 in the ForceAtlas2 paper, https://doi.org/10.1371/journal.pone.0098679). So F_attr ~ 1/W. Here also the display of the edge (line thickness) would have to be modded.

2. Rescale all W in such way, that the ones with largest W become the ones wit smallest W. i.e. idetify edge with largest W (W_max) and smallest W (W_min) and calculate new edge weights as:
W_new = W_max - W_current + W_min
(W_min is added just so the originally strongest edge will not become 0, thus deleted) In this case obviously the original concept of F_attr ~ W will be kept. Here the display of the edge (line thickness) would not need to be modded.

In principle either of the two options could be included as a clickable option that can be selected/de-selected in the option menu for ForceAtlas2 where other parameters are set.

There is currently a pull by astraey to rescale edge weights when negative W is in the graph (https://github.com/gephi/gephi/pull/1733). So he already knows what to modify in case of scenario 2.

Could be useful for many people, since NetworkX has now the option to dump graphs into .gexf format by one command https://networkx.github.io/documentation/stable/reference/readwrite/gexf.html but it obviously writes the weight that are used in NetworkX.
I would really appreciate if this could be solved ;) Thanks'.

## Steps to Reproduce

1.
2.
3.
4.

## Your Environment

* Version used: Gephi 0.9.2
* Java version: 
* Operating System:

pgdr:
## Description

Fix #1816 by implementing inverted edge weight feature in ForceAtlas2 layout engine.

This feature is listed as wishlist 0.10 and in Bug bash #2342 

The change is very small:

```patch
@@ -130,11 +131,12 @@ public class ForceAtlas2 implements Layout {
     }
 
     private double getEdgeWeight(Edge edge, boolean isDynamicWeight, Interval interval) {
-        if (isDynamicWeight) {
-            return edge.getWeight(interval);
-        } else {
-            return edge.getWeight();
-        }
+        double w = edge.getWeight();
+        if (isDynamicWeight)
+            w = edge.getWeight(interval);
+        if (isInvertedEdgeWeightsMode())
+            return w == 0 ? 0 : 1/w;
+        return w;
     }
```

## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md


pgdr:
Adding screenshots of the UI and the difference between inverted and non-inverted using

|    |   source |   target |   weight |
|----|----------|----------|----------|
|  0 |        1 |        2 |        2 |
|  1 |        2 |        3 |       10 |
|  2 |        1 |        4 |        2 |
|  3 |        2 |        4 |        4 |
|  4 |        3 |        4 |       10 |

Normal ForceAtlas2

Inverted Edge Weight ForceAtlas2

pgdr:
@mbastian Is there anything I can do to make the review process easier?

jacomyma (approved):
Reviewed and tested, all fine!

Issue #2736, opened 2023-01-15
Recovering wiped .gephi file caused by Gephi 0.10 bug


ForkedLightning:
Gephi 0.10 overwrote my .gephi file with a blank file yesterday. Does Gephi keep a temporary file containing the data anywhere else for recovery?

I lost a year's worth of work and I don't have access to the original dataset anymore.

Why did the devs push this to release and notified users to upgrade even though such a major bug was known for days? Take it down immediately.

seinecle:
Absolutely sorry for the issue. A fix has been pushed.

PR #2737, opened 2023-01-15
#2736 Fix issue with empty gephi file when save on close

totetmatt:
## Description
When you close gephi, a prompt ask you if you want to save or not your project. If you choose to do so, the resulting gephi is going to be corrupted with only a `Project_xml` inside.

This might be due after investigation to the fact that on exit, gephi is exiting before waiting the end of the save project task.
** Reproducing hte bug **
- Open gephi
- Open les Miserable (or other)
- Merge with current wokrspace (so you have one workspace, otherwise bug will appears on open but it's unrelated #StayTuned  )
- Click on windwos frame button to exit gephi
- Save your file on the "Close Project" Prompt
- Verify that the file written is 400o (or 1k). It can be unzip and only one file will appears.
- Can't be reopen with gephi (or will be empty)

The fix await for the save project task to be finished before continuing.

## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md


modules/DesktopProject/src/main/java/org/gephi/desktop/project/ProjectControllerUIImpl.java
```diff
@@ -281,8 +284,9 @@ public void approveSelection() {
             NbPreferences.forModule(ProjectControllerUIImpl.class).put(LAST_PATH, file.getAbsolutePath());
 
             // Save file
-            saveProject(controller.getCurrentProject(), file);
+            return saveProject(controller.getCurrentProject(), file);
         }
+        return null;
```
mbastian, line 289: Probably better to return `CompletableFuture.completedFuture()`

modules/LongTaskAPI/src/main/java/org/gephi/utils/longtask/api/LongTaskExecutor.java
```diff
@@ -133,12 +133,12 @@ public LongTaskExecutor(boolean doInBackground) {
      *                               <code>taskName</code> is null
      * @throws IllegalStateException if a task is still executing at this time
      */
-    public synchronized void execute(LongTask task, final Runnable runnable, String taskName,
+    public synchronized Future<Object> execute(LongTask task, final Runnable runnable, String taskName,
```
mbastian, line 136: You shouldn't have to change the `LongTaskExectutor`, use the `execute(LongTask task, Callable<V> callable)` method instead

mbastian (changes requested):
Thanks, looking at it I think we should just remove the boolean parameter and always use `true`. It's safer to always wait to save the project before close.

modules/DesktopProject/src/main/java/org/gephi/desktop/project/ProjectControllerUIImpl.java
totetmatt: Done

modules/LongTaskAPI/src/main/java/org/gephi/utils/longtask/api/LongTaskExecutor.java
totetmatt: Ah true ! Changed it to use the Callable version

totetmatt:
Somehow had to change a new Runnable to () ->  in another part of the code. I guess it's ok ?

modules/DesktopProject/src/main/java/org/gephi/desktop/project/ProjectControllerUIImpl.java
```diff
@@ -222,22 +226,28 @@ private void updateTitleBar(Project project) {
...
+            @Override
+            public Void call() throws Exception {
+                controller.saveProject(project, file);
+                return null;
+            }
         });
     }
 
-    public void saveProject() {
+    public Future<Void> saveProject() {
```
totetmatt, line 241: Actually @mbastian Should this function block ? So that in any case we block on save ?

mbastian (approved):
Looks good thank you. I can take another look later to see improvements but as long as it fixes the issue it's looks safe enough.

Issue #2741, opened 2023-01-17
Json position normalization not working


totetmatt:
## Expected Behavior
When selecting on the export option for json "Normalize" Scale position between 0 and 1, the output josn contains normalized positon data
## Current Behavior
Position arn't normalized
## Possible Solution
Make sure the normalization is applied (it wokrs with gexf)

## Steps to Reproduce

1. Open a Graph
2. Save as Json, select Normalize on Option
3. Check that position in the file are not within normalized boundaries

## Your Environment

* Version used: Gephi 0.10.0
* Operating System: Win 10 64 Intel

PR #2742, opened 2023-01-17
Fixing Json export position normalization

totetmatt:
https://github.com/gephi/gephi/issues/2741

## Description
When exporting, you can have the option to normalize the coordinate from 0 to 1.
This PR solve the issue.

Discussion : 
Current behaviour doesn't write the position component (X, Y, Z) in the json if all component = 0 : 
- With the normalization, this behaviour shouldn't appears, therefore added a condition to let write when XYZ=0
- Should we actually write anyway position with component XYZ = 0 i nany case ? 

## Checklist

## Added tests?

- [X] 👍 yes

## Added to documentation?
- [ ] 👍 README.md


mbastian (approved):
Great job!

PR #2758, opened 2023-02-25
First batch (90/226 originals) of the new icons

omroka:
These are all the new icons I have currently made, for both light and dark mode. Original icons from IBM Carbon Design System are under the [Apache 2.0 License ](https://github.com/carbon-design-system/carbon/blob/v10/LICENSE).

Also, I apologize if anything is wrong, this is my first ever pull request.


mbastian:
Hi @omroka, I finally was able to integrate all the new icons you did into Gephi. For now, it's in the development branch but it will be released in the next version. Everything is working smoothly, including the dark mode, which is amazing. Thanks again for the serious work here ❤️. I was wondering if you had another batch in the works? Cheers,

Issue #2842, opened 2023-08-16
Invalid svg polyline points (NaN) exporting curved edges between overlapping nodes.


nathanal:
## Expected Behavior
Valid svg produced on export using [buggy.gephi.zip](https://github.com/gephi/gephi/files/12357134/buggy.gephi.zip)

The edge is between two overlapping nodes.

I'm not sure, either the edge shouldn't be exported, or the polyline should have no points.

## Current Behavior
On export, if two nodes are overlapping and have a shared edge, the path and arrows produced are invalid svg. Specifically the NaN is not valid.

```xml
<path stroke-linecap="round" fill="none" class="id_200 id_1"
              d="M -166.344223,-155.416153 A 2.253508,2.253508 0,0 1,NaN,NaN"
              stroke="#c0c0c0" stroke-width="0.1" stroke-opacity="1.0"/>
<!-- and further along -->
        <polyline fill="#c0c0c0" fill-opacity="1.0" class="id_200 id_1"
                  points="NaN,NaN NaN,NaN NaN,NaN" stroke="none"/>
```

[buggy_trunc](https://github.com/gephi/gephi/assets/22120290/e9a9f889-e900-4a61-80cc-d4a05d055d2b)

Following the SVG 1.1 specs, numbers are expected:

- For polyline: https://www.w3.org/TR/SVG11/shapes.html#PolylineElementPointsAttribute then https://www.w3.org/TR/SVG11/shapes.html#PointsBNF
- For paths: https://www.w3.org/TR/SVG11/paths.html#PathDataBNF

## Possible Solution

Unsure, I'm just starting out the gephi/graphs in general. 

If there shouldn't be any lines, simply having no points  i.e. `points=""` and `d=""` will be fine.

## Steps to Reproduce

1. Open the attached project (it is simplified, only two nodes and an edge) [buggy.gephi.zip](https://github.com/gephi/gephi/files/12357134/buggy.gephi.zip) (rename to buggy.gephi).
2. Go to the preview tab
3. Make sure Show edges, and Curved is checked (under the Edges tab)
4. Export to svg

## Context

Bunch of errors logged in inkscape. I guess it's not that bad. 

## Your Environment

* Version used: Gephi 0.10.1 202301172018
* Operating System: Linux Mint 20 (I'm running in a snap package)

PR #2608, opened 2022-09-02
#1680 Arrows on curved edges

jacomyma:

PR #2844, opened 2023-08-25
#2842 Don't draw invalid svg paths for curved edges

nathanal:
## Description

Fix svg output in #2842 (one node is inside another) by not drawing the edge when it is truncated too much.

Previously, computeThatThing's condition never failed, as rt was only passed when negative. We now pass the arc length and return a large negative angle if it fails. This should register as excessive truncative and be marked not to be drawn.

Used https://github.com/gephi/gephi/pull/2608 and a jumping off point in investigation.

## Checklist

## Added tests?

- [x] 👍 yes

Simple (but big) gexf file to test some cases [testing.zip](https://github.com/gephi/gephi/files/12441889/testing.zip). It got a bit out of hand, but wasn't too smart (just a straight up permutation of various angles, distances, sizes). Smaller version of the above zip [extended_alldir.zip](https://github.com/gephi/gephi/files/12441933/extended_alldir.zip)

Nearly all cases can be considered edge cases (i.e. overlapping nodes).

I've pushed a sample of the above examples cases into a unit test.

## Added to documentation?
- [ ] 👍 README.md

Update: the force push was just squashing my commits together


nathanal:
Thanks you for the review and merge @mbastian !

PR #3001, opened 2025-07-11
Adding functionality to the EgoFilter

floungd:
#### The filter itself:

Gephi has an Ego Network filter which filters the displayed graph according to:

- Starting Node (Ego)
- Depth
- Include Self

After applying said filter, Gephi shows neighbors at depth k, from the "Ego" node.

The Filter does not consider directed edges losing information.

### Desired State:

Have the Ego Network filter consider directed edges, to see nodes connected via outgoing, incoming or any type of edges.

To effectively use the wanted changes 2 major things have to be done:

#### The filter itself:

The Ego network filter now considers the Directed edges and acts according to following rules:

- **Outgoing:** the filter only considers edges that point away from the Ego node.
- **Incoming:** the filter only considers edges that point to the Ego node.
- **Both:** the filter considers incoming and outgoing edges

The filter is able to consider non-directed edges if the user wants it to, keeping the original usability of the filter.

To achieve the desired functionality of the filter some changes must be made to the EgoPanel:

#### The filter panel:

The filter panel now has following options:

- No OK button the filter ego updates on input.
- No depth drop-down the user is able to customize the depth beyond 3 for bigger graphs. to show the maximum amount of edges the field is kept empty.
- Direction drop-down menu to select {Incoming, Outgoing, Both}
- Checkbox for self is kept.
- Checkbox for consideration of non directed graphs. (default true, as in previous behavior)

To achieve this I edited the classes `EgoBuilder` and `EgoPanel`, alongside with the `Bundle.properties`, `Bundle_de.properties`(for JavaSwing names in german) and finally the `egoPanel.form` file due to changes in the UI.

All changes of the UI were done in Apache NetBeans and the code was generated automatically to the `EgoPanel`.

The filter method in `EgoBuilder` was changed entirely, I used a BFS Algorithm to achieve the wanted behavior. 
the method is well documented in the code.

To check whether an edge is relevant the method `isRelevant(Node ego, Edge e)` was implemented, I am using `Enums` to differentiate the options of the added drop-down menu.

##### Default UI:
image

##### Translation:

Knoten ID --> Node ID, Tiefe --> Depth, Richtung --> Direction, Inkl. Eigener --> With Self, berücksichtige ungerichtete Kanten --> Consider non directed Edges

##### Example Graph:

image

NodeId = A || 0, Depth 1, Direction = Outgoing, withSelf = true, considerNonDirectEdges = false

image

NodeId = A || 0, Depth 1, Direction = Incoming, withSelf = true, considerNonDirectEdges = false

image

NodeId = A || 0, Depth 1, Direction = Both, withSelf = true, considerNonDirectEdges = false

image

NodeId = C || 2, Depth 2, Direction = Incoming, withSelf = true, considerNonDirectEdges = true

image

## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md


modules/FiltersPlugin/src/main/java/org/gephi/filters/plugin/graph/EgoBuilder.java
```diff
@@ -108,57 +108,167 @@ public void destroy(Filter filter) {
...
+                    return e.getSource() == ego;
+                case INCOMING:
+                    return e.getTarget() == ego;
+                case BOTH:
+                    return true;
+            }
+            return false;
+        }
+
+        private Node getNextNode(Node ego, Edge e) {
```
mbastian: Redundant with the `graph.getOpposite()` method, consider using that instead, see https://javadoc.io/doc/org.gephi/gephi/latest/org/gephi/graph/api/Graph.html#getOpposite(org.gephi.graph.api.Node,org.gephi.graph.api.Edge)

mbastian (changes requested):
Thanks for the changes, please take a look at my comments

PR #3005, opened 2025-08-18
Adding X and Y Axis control for scaling layout to allow X and Y axis Mirroring

totetmatt:
## Description
Adding boolean option on scaling layout to allow scaling only on X and / or Y axis
(the feature need was expressed by @NicolasBchb during Gephi week )

- xAxis and yAxis boolean parameter + getter and setter added on AbstractScaleLayout
- Created a more generic Scale Layout (@mbastian to discuss if we want to keep actual Expansion / Reduction layout or if it would make sense to make only one "Scale" layout)
## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md


totetmatt:
Might be instead for https://github.com/gephi/gephi/pull/3017

totetmatt:
Creating option for "Expand" and "Contract" layout to apply only operation on one Axis , which allow as a "hack" or "hidden feature" to mirror the graph.

It's a quick implementation to allow user to do the mirror transformation. But it's a temporary solution that might be deprecated in the future with the creation of a proper "TransformationAPI" and "Transformation tab" that will deprecate this kind of layout that doesn't involve any network topology in the layout computation.

PR #3017, opened 2025-08-20
[Gephiweek] POC ToolPlugin for transformation with initial Axes Symetry

totetmatt:
## Description
Creating a Transform Tool plugin that would handle simple graph manipulation. 
Initial operation right now are Vertical Sym and Horizonal Sym .

Might deprecate https://github.com/gephi/gephi/pull/3005 and require design discussion for icon and text.

## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md


totetmatt:
After discussion with @jacomyma , the ToolPlugin is intend to be "action that will involve the mouse" . Therefore the idea is to migrate the set of transofmration tool to another "Tab" like "Spatialisation".
Also, set of action should be as simple as possible so only involving button with predefined value for transformation

totetmatt:
Will create another PR for implementing TransformationAPI

PR #3024, opened 2025-09-05
Transformation Module

totetmatt:
## Description
Initialization of the transformation module that will allow simple graph manipulation like rotation, scale and mirroring .

The idea is to decommission the related Layout to enhance graph manipulation with a whole new Transformation Panel.

Current implementation doesn't allow plugin extension of transformation, but it will come in a future PR, as the code structure needs a proper though, that is slightly different to existing other panel like Layout.
 

## Checklist

## Added tests?

- [X] 👍 yes

## Added to documentation?
- [ ] 👍 README.md

### DIFF
--- a/modules/ToolsPlugin/src/main/java/org/gephi/tools/plugin/EdgePencil.java
+++ b/modules/ToolsPlugin/src/main/java/org/gephi/tools/plugin/EdgePencil.java
@@ -151,6 +151,7 @@ public class EdgePencil implements Tool {
                     boolean directed = edgePencilPanel.isDirected;
                     Edge edge =
                         Lookup.getDefault().lookup(GraphElementsController.class).createEdge(sourceNode, n, directed);
+                    edge.setWeight(weight);
                     edge.setColor(color);
                     sourceNode = null;
                     edgePencilPanel.setStatus(NbBundle.getMessage(EdgePencil.class, "EdgePencil.status1"));
--- a/modules/ToolsPlugin/src/main/java/org/gephi/ui/tools/plugin/EdgePencilPanel.java
+++ b/modules/ToolsPlugin/src/main/java/org/gephi/ui/tools/plugin/EdgePencilPanel.java
@@ -93,7 +93,7 @@ public class EdgePencilPanel extends javax.swing.JPanel {
     }
 
     public void setWeight(float weight) {
-        weightSpinner.getModel().setValue(new Float(weight));
+        weightSpinner.getModel().setValue(weight);
     }
 
     public void setType(boolean directed) {


modules/DesktopTransformation/src/main/nbm/manifest.mf
```diff
@@ -0,0 +1,6 @@
+Manifest-Version: 1.0
+AutoUpdate-Essential-Module: true
+OpenIDE-Module-Localizing-Bundle: org/gephi/desktop/layout/Bundle.properties
```
mbastian, line 3: Typo, this needs to match the package

modules/TransformationAPI/src/main/java/org/gephi/transformation/spi/TransformationOperation.java
```diff
@@ -0,0 +1,19 @@
...
+
+public interface TransformationOperation {
+    void transformation(Graph graph);
+
+    default void apply(Graph graph) {
+        graph.writeLock();
+        try {
+            this.transformation(graph);
+        } catch (Exception e) {
+            graph.writeUnlock();
```
mbastian, line 13: not needed, as finally also gets triggered in case of an exception

modules/TransformationAPI/src/main/nbm/manifest.mf
```diff
@@ -0,0 +1,6 @@
+Manifest-Version: 1.0
+AutoUpdate-Essential-Module: true
+OpenIDE-Module-Localizing-Bundle: org/gephi/layout/api/Bundle.properties
```
mbastian, line 3: This also needs update

modules/TransformationAPI/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>graph-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>project-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
```
mbastian, line 28: I don't think you need this dependency

modules/TransformationPlugin/src/main/java/org/gephi/transformation/plugin/operations/Mirror.java
```diff
@@ -0,0 +1,30 @@
...
+    @Override
+    public void transformation(Graph graph) {
+        float barycenter = 0.f;
+        for (Node n : graph.getNodes()) {
+            barycenter += nodeGetFunction.apply(n);
+        }
+        barycenter /= graph.getNodeCount();
+        for (Node node : graph.getNodes()) {
+            if (node.isFixed()) {
+                return;
```
mbastian, line 24: Looks weird to return here. On top of it, this would NOT release the (auto) read lock, which is a major issue

modules/TransformationPlugin/src/main/java/org/gephi/transformation/plugin/TransformationControllerImpl.java
```diff
@@ -0,0 +1,88 @@
...
+        this.rotation_right_1deg = new Rotation(1.f);
+        this.rotation_left_1deg = new Rotation(-1.f);
+        this.rotation_right_45deg = new Rotation(45.f);
+        this.rotation_left_45deg = new Rotation(-45.f);
+        this.extend = new Scale(1.1f);
+        this.reduce = new Scale(.9f);
+
+    }
+
+    private Graph get_graph() {
```
mbastian, line 41: Convention is to always use camel case in Java

modules/TransformationPlugin/src/main/java/org/gephi/transformation/plugin/TransformationControllerImpl.java
```diff
@@ -0,0 +1,88 @@
...
+import org.gephi.transformation.plugin.operations.MirrorXAxis;
+import org.gephi.transformation.plugin.operations.MirrorYAxis;
+import org.gephi.transformation.plugin.operations.Rotation;
+import org.gephi.transformation.plugin.operations.Scale;
+import org.openide.util.Lookup;
+import org.openide.util.lookup.ServiceProvider;
+
+
+@ServiceProvider(service = TransformationController.class)
+public class TransformationControllerImpl implements TransformationController {
```
mbastian, line 16: On one hand you define a SPI, which is nice because one could write a plugin for new transformation. On the other hand, you hard code what transformations can be executed in the controller. Maybe worth rethinking your approach here.

PR #3045, opened 2025-10-27
GLSL 330

totetmatt:
## Description
Migration to GLSL 330 (mostly updating keyword attribute and varyaring to in and out)

Changing to use a struct for data passed by the vertex shader to the fragment shader.

## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md


eduramiba (approved):
Thanks for the modernization!!

PR #3075, opened 2025-11-20
Feature/screenshot

totetmatt:
## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md

### DIFF
--- a/modules/VisualizationImpl/src/main/java/org/gephi/visualization/screenshot/ScreenshotControllerImpl.java
+++ b/modules/VisualizationImpl/src/main/java/org/gephi/visualization/screenshot/ScreenshotControllerImpl.java
@@ -51,6 +51,8 @@ import org.gephi.viz.engine.jogl.util.Framedata;
 import org.openide.util.Lookup;
 
 import javax.imageio.ImageIO;
+import javax.swing.*;
+import javax.swing.filechooser.FileNameExtensionFilter;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


modules/VisualizationEngine/src/main/java/org/gephi/viz/engine/jogl/JOGLRenderingTarget.java
```diff
@@ -60,6 +73,48 @@ public void setup(VizEngine engine) {
...
+
+
+        ByteBuffer buffer = GLBuffers.newDirectByteBuffer(width * height * 4);
+
+        gl.glReadBuffer(GL_BACK); // Some say GL_FRONT, some say GL_BACK
+
+        gl.glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, buffer);
+
+
+        /* It works, but i'm not fan */
```
eduramiba: I have not tested at all, but GPT5 suggests this:

```java
// Fields you can keep in your renderer class to reuse:
private IntBuffer pixelBuffer = null;
private int[] pixelInts = null;
private BufferedImage screenshot = null;

@Override
public void frameDump(GL3ES3 gl) {
    int width  = drawable.getSurfaceWidth();
    int height = drawable.getSurfaceHeight();
    int size   = width * height;

    // (Re)allocate once if needed
    if (pixelBuffer == null || pixelBuffer.capacity() < size) {
        pixelBuffer = GLBuffers.newDirectIntBuffer(size);
        pixelInts   = new int[size];
    }
    if (screenshot == null ||
        screenshot.getWidth() != width ||
        screenshot.getHeight() != height) {
        screenshot = new BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB);
    }

    gl.glReadBuffer(GL.GL_BACK);
    gl.glPixelStorei(GL.GL_PACK_ALIGNMENT, 1);

    // Read pixels as BGRA packed into an int (fast path for many drivers)
    gl.glReadPixels(
            0, 0, width, height,
            GL.GL_BGRA,
            GL.GL_UNSIGNED_INT_8_8_8_8_REV,
            pixelBuffer
    );

    // Copy from NIO buffer to Java array in one go
    pixelBuffer.rewind();
    pixelBuffer.get(pixelInts);

    // Flip vertically (OpenGL origin is bottom-left, BufferedImage is top-left)
    int[] flipped = new int[size];
    for (int y = 0; y < height; y++) {
        int srcPos  = y * width;
        int dstPos  = (height - 1 - y) * width;
        System.arraycopy(pixelInts, srcPos, flipped, dstPos, width);
    }

    // Fill the BufferedImage in one call
    screenshot.setRGB(0, 0, width, height, flipped, 0, width);

    // TODO: save screenshot or use it as needed
}
```

totetmatt: Ah yeah, we are stuck anyway with the Y flip, but that's what I was looking

PR #3197, opened 2026-05-09
Add ImportException hierarchy for known import failures

mbastian:
## Summary

- Introduce `ImportException` (base) and `EmptyFileException` in the public `ImportAPI` so file importers can signal known, user-facing failures with a localized message, distinguished from genuinely unexpected errors.
- Move the empty-file detection out of individual importers and into `ImportControllerImpl` so every file importer (GEXF, GraphML, GML, GDF, DOT, Pajek, TGF, TLP, DL, VNA, spreadsheets) is protected by the same single check, raised as `EmptyFileException`.
- Teach `DesktopImportControllerUI.ImportErrorHandler` to recognize `ImportException` in the cause chain and surface only the friendly message via the existing issues report panel, instead of the developer-facing "Unexpected Exception" stacktrace popup. Unexpected errors keep their previous behaviour.
- Drop the now-redundant `reader.ready()` check in `ImporterGEXF` (introduced in #3192).
- Mention the new exceptions in the 0.11.0 Import API entry of `overview.html`.

The new types are extensible: future known errors (e.g. unsupported version, malformed root element) can be added as additional `ImportException` subclasses with their own bundle keys, with no further changes required in the controller or the desktop UI.

## Test plan

- [ ] Open an empty `.gexf` file - a friendly "The file '...' is empty." dialog appears in the issues panel, no stacktrace popup.
- [ ] Open an empty `.graphml`, `.gml`, `.gdf`, `.csv` etc. - same friendly dialog now triggered for all file formats.
- [ ] Open a malformed (but non-empty) file - existing developer popup with stacktrace still appears.
- [ ] Open a valid file - import works normally.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/modules/DesktopImport/src/main/java/org/gephi/desktop/importer/DesktopImportControllerUI.java
+++ b/modules/DesktopImport/src/main/java/org/gephi/desktop/importer/DesktopImportControllerUI.java
@@ -67,6 +67,7 @@ import org.gephi.desktop.mrufiles.api.MostRecentFiles;
 import org.gephi.io.importer.api.Container;
 import org.gephi.io.importer.api.Database;
 import org.gephi.io.importer.api.ImportController;
+import org.gephi.io.importer.api.ImportException;
 import org.gephi.io.importer.api.ImportUtils;
 import org.gephi.io.importer.api.Issue;
 import org.gephi.io.importer.api.Report;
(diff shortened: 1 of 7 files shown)


kingmove645-ctrl:
a8be53e5d2302636453de0e5cadd90d4

kingmove645-ctrl:
Viz

kingmove645-ctrl:
Farhan **__**

PR #3218, opened 2026-06-24
Add Windows and macOS to CI build matrix and fix cross-platform issues

addwes7:
## Description
Expand the GitHub Actions CI matrix to run on Windows, macOS, and Ubuntu instead of Ubuntu alone. The project documentation states that Gephi supports Windows, macOS, and Linux. However, the current CI workflow only executes on Ubuntu, which means platform-specific issues on Windows and macOS can go undetected until users encounter them.

### Expected results
The project builds successfully on Ubuntu, Windows, and macOS using JDK 8 while preserving the original build behavior.

### Actual results
In windows, the build fails with the error : `Unknown lifecycle phase ".awt.headless=true"`. This error can be fixed by setting the  MAVEN_OPTS environment variable.

### Description of fix
- Added an OS build matrix (ubuntu-latest, windows-latest, macos-latest).
- Moved Maven build properties (-Djava.awt.headless=true) into MAVEN_OPTS to avoid platform-specific command-line parsing issues.

## Checklist

## Added tests?

- [ ] 👍 yes

## Added to documentation?
- [ ] 👍 README.md

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,12 @@ on:
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
+    runs-on: ${{ matrix.os }}
+    env:
+      MAVEN_OPTS: "-Djava.awt.headless=true"
     steps:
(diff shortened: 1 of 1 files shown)


mbastian:
Thanks for the idea and the fix, you make a valid point!

PR #3236, opened 2026-08-22
Fix NPE when saving a project with an empty text-equals filter value

mbastian:
## Description
When a user adds an "Attributes > Equal" filter for a text column via the Filters panel but never types a value into it before saving the project, the save silently produces broken output today — and on some XML stream implementations, throws an exception in the process.

The root cause is in `EqualStringFilter` (`modules/FiltersPlugin/src/main/java/org/gephi/filters/plugin/attribute/AttributeEqualBuilder.java`): its `pattern` field has no initializer and is never set unless the user calls `setPattern(...)`. Sibling filters like `EgoBuilder.EgoFilter` default their equivalent field to `""`, but this one doesn't, so `pattern` stays `null` for a freshly added, untouched filter.

When the project is saved, `FilterModelPersistenceProvider.writeParameter` (`modules/FiltersImpl/src/main/java/org/gephi/filters/FilterModelPersistenceProvider.java:168`) serializes every filter property by calling `Serialization.toText(value, type)` and passing the result straight to `writer.writeCharacters(...)`. `Serialization.toText` returns `null` for a `null` input value by design, so `writeCharacters(null)` gets called. Depending on the `XMLStreamWriter` implementation in use, that throws an NPE — this is what shows up in Sentry as GEPHI-5SY.

The fix is two-fold and minimal:
- `writeParameter` now guards the serialized text and writes `""` instead of `null`, so no filter property (not just this one) can ever push a null into the XML writer.
- `EqualStringFilter.pattern` now defaults to `""`, matching `EgoFilter`, so the null case doesn't occur in the first place for this filter.

https://gephi.sentry.io/issues/GEPHI-5SY

## Checklist

## Added tests?
- [x] 👍 yes

Added `AttributeEqualBuilderTest.testEqualStringFilterDefaultPatternIsNotNull` (FiltersPlugin), which fails before the fix (asserts the filter built via `AttributeEqualBuilder` never has a null pattern), and a persistence round-trip test in `PersistenceProviderTest` covering an `EqualStringFilter` left without a pattern.

## Added to documentation?

### DIFF
--- a/modules/FiltersImpl/src/main/java/org/gephi/filters/FilterModelPersistenceProvider.java
+++ b/modules/FiltersImpl/src/main/java/org/gephi/filters/FilterModelPersistenceProvider.java
@@ -165,7 +165,8 @@ public class FilterModelPersistenceProvider implements WorkspaceXMLPersistencePr
         try {
             writer.writeStartElement("parameter");
             writer.writeAttribute("index", String.valueOf(index));
-            writer.writeCharacters(serialization.toText(property.getValue(), property.getValueType()));
+            String text = serialization.toText(property.getValue(), property.getValueType());
+            writer.writeCharacters(text != null ? text : "");
             writer.writeEndElement();
         } catch (Exception e) {
             Exceptions.printStackTrace(e);
--- a/modules/FiltersImpl/src/test/java/org/gephi/filters/PersistenceProviderTest.java
+++ b/modules/FiltersImpl/src/test/java/org/gephi/filters/PersistenceProviderTest.java
@@ -7,6 +7,7 @@ import org.gephi.filters.plugin.graph.HasSelfLoopBuilder;
 import org.gephi.filters.plugin.operator.INTERSECTIONBuilder;
 import org.gephi.filters.spi.Filter;
 import org.gephi.filters.spi.FilterBuilder;
+import org.gephi.graph.api.Column;
 import org.gephi.project.io.utils.GephiFormat;
 import org.junit.Assert;
 import org.junit.Ignore;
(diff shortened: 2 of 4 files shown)


mbastian:
The fix looks correct and minimal for the reported crash (`EqualStringFilter.pattern` defaulting to `""` plus the null-guard in `FilterModelPersistenceProvider.writeParameter`), and the round-trip test correctly exercises the previously-broken path.

However, there's an identical unguarded call site that this PR doesn't touch: `LayoutModelImpl.writeXML` (`modules/LayoutAPI/src/main/java/org/gephi/layout/LayoutModelImpl.java:311`):

```java
for (Entry<LayoutPropertyKey, Object> entry : savedProperties.entrySet()) {
    if (entry.getValue() != null) {
        ...
        writer.writeCharacters(Serialization.getValueAsText(entry.getValue()));
        ...
```

The `entry.getValue() != null` check only guards against a null *value*, but `Serialization.toText(...)` (which `getValueAsText` delegates to) can still return `null` for a non-null value whenever `PropertyEditorManager.findEditor(valueClass)` finds no registered editor for that property's type (see the final `else { return null; }` branch in `Serialization.toText`). That happens for any layout property whose type isn't a primitive/wrapper/Number/array/Color/Font/String and has no `PropertyEditor` registered — e.g. a custom or enum-typed property from a third-party layout plugin — which would hit `writeCharacters(null)` and NPE exactly like GEPHI-5SY did for filters.

Unlike `writeParameter`, which swallows exceptions in a `catch (Exception e)` block, `LayoutModelPersistenceProvider.writeXML` only catches `XMLStreamException`, so this NPE would propagate uncaught on project save. Given the PR's stated goal ("no filter property... can ever push a null into the XML writer"), it might be worth applying the same guard here too, since it's the same underlying `Serialization` null-propagation pattern in a sibling persistence provider.

PR #3237, opened 2026-08-22
Null-check graph model/edge before use in async pencil-tool click dispatch

mbastian:
## Description

`VizEngine.scheduleNextConcurrentWorldUpdateIfDone` (`modules/VisualizationEngine/src/main/java/org/gephi/viz/engine/VizEngine.java:721-748`) queues each mouse click made with the Node/Edge pencil tools onto a background executor as part of the async world-update pipeline; the click is actually dispatched later, inside `processInputEvents`, from the `CompletableFuture` task (`VizEngine.java:663`, `VizEngine.java:866`). By the time that task runs, the graph the user was pointing at may no longer be in the same state it was when the click was captured — the workspace can have been closed, or a node involved in the click can have been removed (e.g. by an undo or a concurrent edit).

Both pencil tools assumed their graph lookups always succeed and dereferenced the result unconditionally:

- `NodePencil` (`NodePencil.java:102-103`) calls `GraphController.getGraphModel()`, which is documented to return `null` "if project is empty", then immediately calls `gm.getGraph()`. When the async click runs after the project has emptied out, this NPEs on the render thread (GEPHI-5HN, 59 users / 552 events).
- `EdgePencil` (`EdgePencil.java:155-157`) calls `GraphElementsController.createEdge(...)`, documented to return "New edge if the edge was created succesfully, null otherwise", then immediately calls `edge.setWeight(...)`. When the async click's source/target node is no longer valid in the graph store, `createEdge` catches the internal failure and returns `null`, and the very next line NPEs (GEPHI-5MZ, 7 users / 69 events).

Both crashes surface as an `ExecutionException`/`GLException` wrapping the NPE, unwound from `VizEngine.runWorldUpdated` on the render thread, which is why the visible symptom looks like a rendering-engine crash even though the root cause is in the tool code.

The fix adds the null checks the documented contracts already call for: if `getGraphModel()` returns `null`, `NodePencil` no-ops the click instead of dereferencing it; if `createEdge(...)` returns `null`, `EdgePencil` resets its pending-source-node state and no-ops instead of dereferencing the edge. This doesn't change behavior for any successful click and matches the existing return-`false`/"nothing happened" convention already used elsewhere in these listeners (e.g. `EdgePencil`'s cancel-click case). A more thorough fix would cancel/guard pending world-update tasks when a workspace closes, but that's a larger change to `VizEngine`'s scheduling and out of scope for this crash fix.

Sentry: https://gephi.sentry.io/issues/GEPHI-5HN, https://gephi.sentry.io/issues/GEPHI-5MZ

## Checklist

## Added tests?
- [x] 👍 yes

Added `NodePencilTest` and `EdgePencilTest` (new test sources for this module). Each reproduces the exact NPE from the corresponding Sentry stack trace against the pre-fix code (verified locally by reverting the fix and confirming both tests fail with the same NPE messages: `"gm" is null` at `NodePencil.java:103` and `"edge" is null` at `EdgePencil.java:157`), then asserts the click is safely no-op'd after the fix:
- `NodePencilTest` calls the tool's mouse-click listener with no project open, so `GraphController.getGraphModel()` returns `null`.
- `EdgePencilTest` clicks a source node, then removes that node from the graph before the second click, so `GraphElementsController.createEdge(...)` fails internally and returns `null`.

## Added to documentation?

### DIFF
--- a/modules/ToolsPlugin/src/main/java/org/gephi/tools/plugin/EdgePencil.java
+++ b/modules/ToolsPlugin/src/main/java/org/gephi/tools/plugin/EdgePencil.java
@@ -154,6 +154,11 @@ public class EdgePencil implements Tool {
                     boolean directed = edgePencilPanel.isDirected;
                     Edge edge =
                         Lookup.getDefault().lookup(GraphElementsController.class).createEdge(sourceNode, n, directed);
+                    if (edge == null) {
+                        sourceNode = null;
+                        edgePencilPanel.setStatus(NbBundle.getMessage(EdgePencil.class, "EdgePencil.status1"));
+                        return false;
+                    }
                     edge.setWeight(weight);
                     edge.setColor(color);
                     sourceNode = null;
--- a/modules/ToolsPlugin/src/main/java/org/gephi/tools/plugin/NodePencil.java
+++ b/modules/ToolsPlugin/src/main/java/org/gephi/tools/plugin/NodePencil.java
@@ -100,6 +100,9 @@ public class NodePencil implements Tool {
                 size = nodePencilPanel.getNodeSize();
                 GraphController gc = Lookup.getDefault().lookup(GraphController.class);
                 GraphModel gm = gc.getGraphModel();
+                if (gm == null) {
+                    return false;
+                }
                 Graph graph = gm.getGraph();
                 Node node = gm.factory().newNode();
                 node.setX(position3d[0]);
--- a/modules/ToolsPlugin/src/test/java/org/gephi/tools/plugin/EdgePencilTest.java
+++ b/modules/ToolsPlugin/src/test/java/org/gephi/tools/plugin/EdgePencilTest.java
@@ -1,0 +1,87 @@
+/*
+ Copyright 2008-2010 Gephi
+ Authors : Mathieu Bastian <mathieu.bastian@gephi.org>
+ Website : http://www.gephi.org
+
+ This file is part of Gephi.
+
+ DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+
+ Copyright 2011 Gephi Consortium. All rights reserved.
+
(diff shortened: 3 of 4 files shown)


mbastian:
The null checks in `NodePencil`/`EdgePencil` look correct and are confirmed against the actual Sentry stack traces (GEPHI-5HN NPEs at exactly `NodePencil.java:103` on `gm.getGraph()`, GEPHI-5MZ NPEs at exactly `EdgePencil.java:157` on `edge.setWeight(...)`).

However, the same defect class is present in at least two other tools in this module that go through the identical async dispatch path (`VizEngine.processInputEvents` → `scheduleNextConcurrentWorldUpdateIfDone`), and weren't touched here:

- `ShortestPath.java:124-130` — `clickNodes(...)` (a `NodeClickEventListener`, same dispatch mechanism as `EdgePencil`) does `GraphModel gm = gc.getGraphModel();` then immediately calls `gm.isDirected()` with no null check and no try/catch, so if the workspace closes while a shortest-path click is queued, it will NPE uncaught on the render thread the same way GEPHI-5HN/5MZ did.
- `Brush.java:120` — `getDiffusedNodes(...)` (called from `pressingNodes(...)`, a `NodePressingEventListener` on the same event pipeline) does `GraphModel model = Lookup.getDefault().lookup(GraphController.class).getGraphModel();` then dereferences `model.getGraphVisible()`/`model.isDirected()` unconditionally, also unguarded and uncaught.

(`HeatMap.java:122` has the same unguarded `getGraphModel()` call but it's already wrapped in a catch-all `try/catch (Exception e)`, so it won't crash — just worth noting it's not actually affected.)

Since the PR's own rationale for bundling NodePencil+EdgePencil together is "same defect class... worth fixing together since the guard is the same shape in both places," it'd be worth adding the same guard to `ShortestPath` and `Brush` here too, rather than leaving two more Sentry crashes waiting to happen behind the same root cause.

PR #3249, opened 2026-08-23
Report statistics execution failures with the failing algorithm name

mbastian:
## Description

When a statistics algorithm throws, the user gets an anonymous "unexpected exception" crash report, no indication of which algorithm failed, and a Statistics panel row that keeps spinning for the rest of the session — 16 users and 48 events, all from a third-party "Bridging Centrality" plugin whose metric computation feeds a NaN value into `new BigDecimal(...)`.

**Mechanism**

* `StatisticsControllerUIImpl.execute()` (`modules/DesktopStatistics/src/main/java/org/gephi/desktop/statistics/StatisticsControllerUIImpl.java:149`) marks the statistics as running (`StatisticsModelUIImpl.setRunning(statistics, true)`) and calls `StatisticsControllerImpl.execute(Statistics, LongTaskListener)` (`modules/StatisticsAPI/src/main/java/org/gephi/statistics/StatisticsControllerImpl.java:106`).
* That method creates its own executor (`StatisticsControllerImpl.java:108`) and sets a `LongTaskListener` on it, but never sets an error handler, and both `executor.execute(...)` calls pass `null` as the trailing error-handler argument (`StatisticsControllerImpl.java:122` and `:131`).
* The algorithm runs inside `LongTaskExecutor.RunningLongTask.call()`, whose `catch (Throwable e)` (`modules/LongTaskAPI/src/main/java/org/gephi/utils/longtask/api/LongTaskExecutor.java:364`) tries the per-task handler, then the executor's default handler, and — both being `null` here — falls through to the last-resort `Logger.getLogger("").log(Level.SEVERE, "", e)` (`LongTaskExecutor.java:375`).
* That log record carries no message, so the raw throwable is all that is left: the root-logger handler installed at `modules/DesktopBranding/src/main/java/org/gephi/branding/desktop/Installer.java:126` picks it up in `ReporterHandler.publish()` and turns it into a Gephi crash report (`ReporterHandler.java:116`), with nothing telling the user that a statistics algorithm — here a plugin Gephi has no control over — is what failed.
* The failing task is then completed with `finished(this, false)` (`LongTaskExecutor.java:383`), i.e. without notifying the listener — failure is the only completion path that skips it, cancellation notifies via `InterruptTimerTask` (`LongTaskExecutor.java:457`). So `StatisticsControllerUIImpl`'s listener never runs, the statistics stays in `StatisticsModelUIImpl`'s running list (`StatisticsModelUIImpl.java:118`), and `StatisticsFrontEnd.refreshModel()` keeps that row busy with a "Cancel" button that only calls `statistics.cancel()` and can never clear the flag. `StatisticsUI.unsetup()` is skipped as well.

<details>
<summary>Stack trace (GEPHI-5GK)</summary>

```
at java.lang.Thread.run(Unknown Source)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
at java.util.concurrent.FutureTask.run(Unknown Source)
at org.gephi.utils.longtask.api.LongTaskExecutor$RunningLongTask.call(LongTaskExecutor.java:336)
at org.gephi.statistics.StatisticsControllerImpl$2.run(StatisticsControllerImpl.java:129)
at org.gephi.statistics.StatisticsControllerImpl.executeStatic(StatisticsControllerImpl.java:148)
at ia.ppgco.facom.ufu.br.gephi.plugin.bridgingcoefficient.BridgingCentralityMetric.execute(BridgingCentralityMetric.java:89)
at ia.ppgco.facom.ufu.br.gephi.plugin.bridgingcoefficient.BridgingCentralityMetric.execute(BridgingCentralityMetric.java:117)
at ia.ppgco.facom.ufu.br.gephi.plugin.bridgingcoefficient.BridgingCentralityMetric.saveCalculatedValues(BridgingCentralityMetric.java:321)
at java.math.BigDecimal.<init>(Unknown Source)
at java.math.BigDecimal.<init>(Unknown Source)
```

</details>

**What changed**

`StatisticsControllerImpl.execute(Statistics, LongTaskListener)` now sets a default error handler on the executor it creates — the piece every other long-task user already has (`LayoutModelImpl`, `DesktopGeneratorController`, `DesktopExportController`, `DesktopImportControllerUI`). The handler:

1. reports the throwable with `Exceptions.printStackTrace(t)`, unmodified, so the existing crash report and its Sentry grouping are untouched;
2. shows a `NotifyDescriptor.Message` naming the statistics that failed and the error it failed with, so the user learns which algorithm gave up — same way `StatisticsFrontEnd` already tells the user a statistics can't run;
3. notifies the `LongTaskListener` (which is right there in scope), so the Statistics panel leaves the running state and the row can be run again.

An earlier revision of this PR attached the message to the throwable with `Exceptions.attachLocalizedMessage()` instead. That only reached the log annotation: the localized text is stored in a `LogRecord` whose message is the `"msg"` placeholder, so `AnnException.getMessage()` stays empty, `ReporterHandler.createMessage()` walks past the zero-frame annotation and still reports `NumberFormatException: Infinite or NaN`, and the throwable passed to `Sentry.captureException()` only gained an empty extra cause. Hence the plain dialog and an untouched throwable.

This cannot fix the plugin's NaN computation and does not try to: the report stays (a plugin crashing for 16 users is worth seeing, and the stack trace already names the plugin), but the failure is no longer anonymous to the user and no longer leaves the panel stuck.

Two sibling call sites still pass no error handler to their executor — `ScreenshotControllerImpl.takeScreenshot()` (VisualizationImpl) and `ProjectControllerImpl`'s `longTaskExecutor` (ProjectAPI). Left out of this PR.

Sentry issue: https://gephi.sentry.io/issues/GEPHI-5GK

## Checklist

## Added tests?

- [x] 👍 yes

`StatisticsControllerImplTest` runs a statistics that throws `NumberFormatException("Infinite or NaN")` through `execute(Statistics, LongTaskListener)`, with a mock `DialogDisplayer` in the lookup, and asserts that the failure is logged, that the message shown to the user names the statistics, and that the listener is notified. Each assertion fails without the corresponding part of the fix.

## Added to documentation?
- [ ] 👍 README.md

### DIFF
--- a/modules/StatisticsAPI/pom.xml
+++ b/modules/StatisticsAPI/pom.xml
@@ -38,6 +38,10 @@
         </dependency>
         <dependency>
             <groupId>org.netbeans.api</groupId>
+            <artifactId>org-openide-dialogs</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.netbeans.api</groupId>
             <artifactId>org-openide-util-lookup</artifactId>
         </dependency>
         <dependency>
--- a/modules/StatisticsAPI/src/main/java/org/gephi/statistics/StatisticsControllerImpl.java
+++ b/modules/StatisticsAPI/src/main/java/org/gephi/statistics/StatisticsControllerImpl.java
@@ -63,6 +63,8 @@ import org.gephi.utils.longtask.api.LongTaskListener;
 import org.gephi.utils.longtask.spi.LongTask;
 import org.gephi.utils.progress.Progress;
 import org.gephi.utils.progress.ProgressTicket;
+import org.openide.DialogDisplayer;
+import org.openide.NotifyDescriptor;
 import org.openide.util.Exceptions;
 import org.openide.util.Lookup;
 import org.openide.util.NbBundle;
@@ -112,9 +114,14 @@ public class StatisticsControllerImpl implements StatisticsController, Controlle
             executor.setLongTaskListener(listener);
         }
         executor.setDefaultErrorHandler(t -> {
-            String message = NbBundle.getMessage(StatisticsControllerImpl.class,
-                "StatisticsControllerImpl.errorHandler.critical", builder.getName(), t.toString());
-            Exceptions.printStackTrace(Exceptions.attachLocalizedMessage(t, message));
+            Exceptions.printStackTrace(t);
+            String cause = t.getLocalizedMessage() != null ? t.getLocalizedMessage() : t.toString();
+            DialogDisplayer.getDefault().notifyLater(new NotifyDescriptor.Message(NbBundle
+                .getMessage(StatisticsControllerImpl.class, "StatisticsControllerImpl.errorHandler.critical",
+                    builder.getName(), cause), NotifyDescriptor.ERROR_MESSAGE));
+            if (listener != null) {
+                listener.taskFinished(statistics instanceof LongTask ? (LongTask) statistics : null);
+            }
         });
 
         final StatisticsModelImpl model = getModel();
--- a/modules/StatisticsAPI/src/main/java/org/gephi/statistics/StatisticsControllerImpl.java
+++ b/modules/StatisticsAPI/src/main/java/org/gephi/statistics/StatisticsControllerImpl.java
@@ -63,7 +63,9 @@ import org.gephi.utils.longtask.api.LongTaskListener;
 import org.gephi.utils.longtask.spi.LongTask;
 import org.gephi.utils.progress.Progress;
 import org.gephi.utils.progress.ProgressTicket;
+import org.openide.util.Exceptions;
 import org.openide.util.Lookup;
+import org.openide.util.NbBundle;
 import org.openide.util.lookup.ServiceProvider;
 import org.openide.util.lookup.ServiceProviders;
 
@@ -104,11 +106,16 @@ public class StatisticsControllerImpl implements StatisticsController, Controlle
 
     @Override
     public void execute(final Statistics statistics, LongTaskListener listener) {
-        StatisticsBuilder builder = getBuilder(statistics.getClass());
+        final StatisticsBuilder builder = getBuilder(statistics.getClass());
         LongTaskExecutor executor = new LongTaskExecutor(true, "Statistics " + builder.getName(), 10);
         if (listener != null) {
             executor.setLongTaskListener(listener);
         }
+        executor.setDefaultErrorHandler(t -> {
+            String message = NbBundle.getMessage(StatisticsControllerImpl.class,
+                "StatisticsControllerImpl.errorHandler.critical", builder.getName(), t.toString());
+            Exceptions.printStackTrace(Exceptions.attachLocalizedMessage(t, message));
+        });
 
         final StatisticsModelImpl model = getModel();
         if (statistics instanceof DynamicStatistics) {
--- a/modules/StatisticsAPI/src/main/resources/org/gephi/statistics/Bundle.properties
+++ b/modules/StatisticsAPI/src/main/resources/org/gephi/statistics/Bundle.properties
@@ -1,0 +1,1 @@
+StatisticsControllerImpl.errorHandler.critical={0} failed to execute due to the following error: {1}
--- a/modules/StatisticsAPI/src/test/java/org/gephi/statistics/StatisticsControllerImplTest.java
+++ b/modules/StatisticsAPI/src/test/java/org/gephi/statistics/StatisticsControllerImplTest.java
@@ -1,6 +1,9 @@
 package org.gephi.statistics;
 
+import java.awt.Dialog;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Handler;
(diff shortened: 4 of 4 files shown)


mbastian:
I checked the mechanism rather than the shape, and the reporting half holds up: `LongTaskExecutor.RunningLongTask.call()`'s catch really does fall through to `Logger.getLogger("").log(Level.SEVERE, "", e)` (`LongTaskExecutor.java:375`) when both the per-task and the default handler are null, `setDefaultErrorHandler` is consulted at `:372`, and because the handler is installed on the executor instead of passed per call it covers the dynamic branch as well as the static one. The new test is a real regression test — I dropped `master`'s `StatisticsControllerImpl` back in and it fails with `AssertionError: The reported failure has no message for the user`, and it passes on the branch. `-P enableCheckStyle` is clean on the module, and the new bundle doesn't disturb the generated module manifest (`OpenIDE-Module-Localizing-Bundle` still points at `org/gephi/statistics/api/Bundle.properties`).

Two things I'd push back on.

### 1. The stuck Statistics panel is fixable here, not only in DesktopStatistics

The listed out-of-scope item ("`LongTaskExecutor` doesn't notify the listener on failure") is accurate as a description — `call()` does `finished(this, false)` in the `finally` — but it means this PR leaves the symptom a user actually notices. `StatisticsControllerUIImpl.execute()` sets `model.setRunning(statistics, true)` and clears it only from `taskFinished`, so after a failure `StatisticsFrontEnd.refreshModel()` keeps the busy spinner and leaves the button as "Cancel" indefinitely. `StatisticsFrontEnd.cancel()` only calls `statistics.cancel()` and never clears the running flag, so that row can't be re-run for the life of the workspace, and `StatisticsUI.unsetup()` never runs either.

That doesn't require touching DesktopStatistics or LongTaskExecutor: `listener` is in scope and effectively final in the method being edited, so the handler added at `StatisticsControllerImpl.java:114-118` can close it:

```java
executor.setDefaultErrorHandler(t -> {
    String message = NbBundle.getMessage(...);
    Exceptions.printStackTrace(Exceptions.attachLocalizedMessage(t, message));
    if (listener != null) {
        listener.taskFinished(statistics instanceof LongTask ? (LongTask) statistics : null);
    }
});
```

Worth noting that cancellation *does* notify the listener (`InterruptTimerTask.run()` -> `finished(task, true)`), so failure is the only completion path that leaves the UI in the running state — which reads more like an oversight in the executor than a deliberate contract. Fixing it here or filing the follow-up both work, but as it stands someone who hits GEPHI-5GK gets a better dialog plus a permanently spinning row.

### 2. The algorithm name doesn't reach the crash report or the Sentry event, only the log attachment

`attachLocalizedMessage` stores the text in a `LogRecord` whose message is the literal `"msg"` placeholder, resolved via an attached `ResourceBundle`, and `AnnException.getMessage()` deliberately skips records whose message equals `"msg"`. So the annotation's own `toString()` is empty, and on the Gephi side:

- `ReporterHandler.createMessage()` walks past zero-frame causes and returns `thr.toString()`, so `Report.summary` stays `NumberFormatException: Infinite or NaN` — unchanged.
- `ReportController.captureExceptionToSentry()` calls `Sentry.captureException(report.getThrowable(), hint)`, and that throwable now carries an extra cause (`Exceptions$AnnException`) with an empty message and no frames. The Sentry title/culprit are therefore no more informative than before, and the exception chain has one more link than it used to — probably worth checking whether that re-fingerprints GEPHI-5GK into a new issue.
- The text *does* land in the attached `messages.log`, because `NbFormatter.extractDelegates()` picks up the `Callable<LogRecord[]>` annotation and formats its record through the bundle. (That isn't visible in surefire output, which uses `SimpleFormatter` — hence the bare `Caused by:` with nothing after it in the test log.)

If the intent is for the report itself to name the algorithm, adding `Exceptions.attachMessage(t, message)` alongside the localized one puts the text into `AnnException.getMessage()` and hence into the Sentry exception value; that's the usual NetBeans pairing (`attachMessage` for the log, `attachLocalizedMessage` for the dialog). Also a nit: `{1}` is `t.toString()`, so the dialog reads "... due to the following error: java.lang.NumberFormatException: Infinite or NaN"; `t.getLocalizedMessage()` would read better.

Not blocking, just for the record — the same `null`-error-handler fallback is still present in `ScreenshotControllerImpl.takeScreenshot()` (VisualizationImpl) and on `ProjectControllerImpl`'s `longTaskExecutor` (ProjectAPI). Reasonable to leave them out of this PR, but those are the remaining two.

mbastian:
Both points were right, and both are fixed in 56a3b3d.

**The localized message never reached the report.** I reproduced it outside Gephi against `org-openide-util` RELEASE290: `attachLocalizedMessage(t, text)` alone leaves `t.getCause().getMessage()` empty, so `ReporterHandler.createMessage()` — which walks past zero-frame causes — still returns `NumberFormatException: Infinite or NaN`, and `ReportController.captureExceptionToSentry()` hands `Sentry.captureException()` a throwable with one extra empty link. So the only thing the first revision really improved was the `messages.log` attachment.

Rather than pair it with `attachMessage`, the handler now shows a `NotifyDescriptor.Message` naming the statistics and the error it failed with, and reports the throwable **unmodified**. That way the message is in something the user actually reads, and the exception chain handed to Sentry is byte-identical to master's, so GEPHI-5GK keeps grouping instead of re-fingerprinting. It also matches how `StatisticsFrontEnd` already tells the user a statistics can't run. This adds an `org-openide-dialogs` dependency to StatisticsAPI (`DataLaboratoryAPI` does the same); `DialogDisplayer.Trivial.notify()` returns `CLOSED_OPTION` immediately when `GraphicsEnvironment.isHeadless()`, so a headless toolkit run is unaffected — and the only caller of this `execute` overload is the desktop UI anyway. Nit taken too: the error text now comes from `t.getLocalizedMessage()`, falling back to `toString()` so it can't read "... error: null".

**The stuck panel was in scope.** You're right that `listener` is sitting there effectively final, and confirmed on the UI side: `StatisticsModelUIImpl.setRunning()` only removes from `runningList` on `taskFinished`, so the row stays busy with a "Cancel" that calls `statistics.cancel()` and can never clear the flag. The handler now calls `listener.taskFinished(...)` after reporting, in that order so a listener that throws can't swallow the report.

The test grew two assertions to match — a mock `DialogDisplayer` in the lookup asserts the user-visible message names the statistics, and a latch asserts the listener fires. I checked each one fails on its own: unwiring the handler fails with "The statistics failure isn't reported to the user", and keeping the handler but dropping the notification fails with "The listener isn't notified of the failure".

`ScreenshotControllerImpl` and `ProjectControllerImpl` left alone as agreed.

Issue #3246, opened 2026-08-23
Audit: WorkspaceListener delivery races on an ambiguous thread (EDT vs. background) — 11 findings


mbastian:
## Background

PR #3232 fixed a Sentry-reported NPE in `AttributesUIControllerImpl`: `ProjectController` fires `WorkspaceListener` callbacks (`initialize`/`select`/`unselect`/`close`/`disable`) synchronously on whatever thread called `openWorkspace()`/`closeCurrentWorkspace()`/etc — sometimes the EDT, sometimes a background thread (e.g. during project loading) — with no dispatch guarantee. A listener registered mid-construction on the EDT raced against a background-thread-fired event and read a field that hadn't been assigned yet.

Following that fix, we audited every `WorkspaceListener` implementer in the codebase for the same bug class. This issue tracks the findings.

## Structural fact that constrains any fix

`fireWorkspaceEvent` (`modules/ProjectAPI/src/main/java/org/gephi/project/impl/ProjectControllerImpl.java:482`) runs inside `synchronized (this)` on the controller for every call site (`openWorkspace`, `closeCurrentWorkspace`, `closeCurrentProject`, etc.). That monitor currently:
- Is the only thing serializing the event sequence and giving several listeners' non-`volatile` field writes a happens-before edge (some are correct today only by accident of this lock).
- Means any listener calling `SwingUtilities.invokeAndWait` while the EDT is blocked on any synchronized `ProjectController` method deadlocks (already the subject of a comment on `getCurrentWorkspace()`).
- Must be addressed (fire moved outside the lock, `CLOSE`/`UNSELECT` semantics decided explicitly) **before** any move to deterministic/async delivery — otherwise a `RequestProcessor`-based redesign just relocates the deadlock risk.

## Live, reproducible bugs (fixable now, independent of any redesign)

1. **`DataTableTopComponent.java:729-735` / `:1603`** — a raw background `Thread` (`RefreshOnceHelperThread`) calls Swing setters (`bannerPanel.setVisible`, `nodesButton.setSelected`, `edgesButton.setSelected`) with zero EDT guard. Reached via `select()` → `activateWorkspace()` → `refreshAllOnce()`. Also: `dataTablesObservers` field is not `volatile`, double-read in a timer task → possible NPE race against `unselect()`.
[...]

PR #3252, opened 2026-08-24
#3246 Close listener-registration races in the Appearance and Preview controllers

mbastian:
## Description

Addresses findings 3 and 4 of #3246. Both are instances of the same bug class fixed in `AttributesUIControllerImpl`/`AttributesTopComponent` by PR #3232: `ProjectControllerImpl.fireWorkspaceEvent` delivers `WorkspaceListener` callbacks synchronously on whatever thread called `openWorkspace()`/`closeCurrentWorkspace()`/etc, sometimes the EDT, sometimes a background thread during project loading.

**Finding 3 — `AppearanceUIController`**

`listeners` was declared `final` at `AppearanceUIController.java:78` but only assigned at line 133, while `pc.addWorkspaceListener(...)` registered the anonymous `WorkspaceListener` at line 85, 48 lines and a `Lookup` call earlier in the same constructor. Any workspace event landing in that window reached `firePropertyChangeEvent`, which iterates `listeners`, and NPE'd on the still-unassigned final field. The same window left `transformers` (assigned at line 135) null.

`select()` (line 101) and `disable()` (line 119) both called `firePropertyChangeEvent` directly, with no thread guard. The only two implementations of `AppearanceUIModelListener`, `AppearanceTopComponent` and `AppearanceToolbar`, are plain Swing and don't guard their `propertyChange` entry point, so an event fired from a background thread ran their Swing-touching code off the EDT.

Fix: `listeners` is now initialized at its declaration. `pc.addWorkspaceListener(...)` is registered last, after `transformers` is populated, with a comment explaining why. `model` is `volatile`. `select()` and `disable()` snapshot `oldModel`/the new model into locals and defer `firePropertyChangeEvent` through a `runAction` helper (inline if already on the EDT, `SwingUtilities.invokeLater` otherwise) so listener fan-out always lands on the EDT, while the model mutation itself (`model.select()`, the field assignment) stays on the calling thread, matching `AttributesUIControllerImpl`'s pattern. `select()` now reuses the existing `getModel(Workspace)` helper instead of duplicating its lookup-or-create logic inline, which also gives the lambda a single-assignment local to capture.

**Finding 4 — `PreviewUIControllerImpl` and its three second-layer listeners**

`listeners` was a plain `ArrayList` (`PreviewUIControllerImpl.java:100`), mutated on the EDT by `addPropertyChangeListener` (called from three TopComponent/panel constructors) and iterated by `fireEvent` from whatever thread fires `select()`/`unselect()`/`disable()`, an unsynchronized read/write race with a real `ConcurrentModificationException` window. `pc.addWorkspaceListener(...)` was registered at line 108, before the `PropertyEditorManager.registerEditor(...)` calls that finish the constructor at line 152. `select()`/`unselect()`/`disable()` fired events with no dispatch guard, same as finding 3.

Fix: `listeners` is now a `CopyOnWriteArrayList` (keeps the existing `contains` guard and ordering in `addPropertyChangeListener`). `pc.addWorkspaceListener(...)` is registered last. `select()`/`unselect()`/`disable()` defer their `fireEvent` call through the same `runAction` pattern, snapshotting the model into a local first. The `REFRESHING`/`REFRESHED` events fired from the `"Refresh Preview"` background thread in `refreshPreview()` are untouched, since that path is a separate mechanism from workspace-event delivery. `PreviewTopComponent` already defers both of those branches to the EDT. `PreviewSettingsTopComponent` does not: its `REFRESHING` branch calls `enableRefreshButton()`/`disableRefreshButton()` inline, and those are unguarded `setEnabled()` calls, so every Refresh click touches Swing off the EDT. That is a real bug, but it is not a `WorkspaceListener` delivery race, so it is tracked as finding 11 of #3246 rather than widened into this PR.

The three second-layer listeners each registered before finishing their own constructor: `PreviewTopComponent.java:169` before assigning `this.model`/calling `initTarget(model)` (172-173); `PreviewSettingsTopComponent.java:207` before `setup(m)`/`enableRefreshButton()` (211-212); `RendererManager.java:109` before `panel.setLayout(...)`/`setup()` (110-111). All three are Swing components built by the NetBeans window system (`RendererManager` itself is instantiated inside `PreviewSettingsTopComponent`'s own constructor), so their construction always runs on the EDT, and registering the listener no longer risks a synchronous cross-thread delivery once `PreviewUIControllerImpl` only ever delivers `SELECT`/`UNSELECT` on the EDT. I moved each `addPropertyChangeListener(this)` call to the end of its constructor anyway, matching the hardening PR #3232 applied to `AttributesTopComponent`: it costs nothing (pure reordering, no behavior change), and it removes reliance on the unenforced assumption that these components are always constructed on the EDT.

## Checklist

## Added tests?

- [x] 👍 yes

Added `AppearanceUIControllerTest` covering finding 3: one test constructs the controller and immediately drives a real workspace-select event through it to confirm the listener set and transformer registry are usable right away; the other starts a plain background `Thread` that calls `ProjectController.openWorkspace(...)` and asserts the registered listener's `propertyChange` only runs once `SwingUtilities.isEventDispatchThread()` is true, never on the firing thread. I verified this second test fails against the pre-fix code (reverting the `runAction` wrap makes it fail deterministically) before confirming it passes against the fix.

I did not add a test for finding 4. `PreviewUIControllerImpl`'s workspace-select path unconditionally builds a `PreviewUIModelImpl`, which needs a real `PreviewController`/`GraphModel`-backed workspace to construct without throwing, and `PreviewSettingsTopComponent`/`PreviewTopComponent`/`RendererManager` are NetBeans `TopComponent`/`JPanel` subclasses that pull in Swing form-generated fields and can't be constructed standalone in a headless unit test without the window system. The fix mirrors the pattern validated by the Appearance test above (same `runAction` helper, same registration reordering), so I judged writing a second, weaker test not worth the added test-only scaffolding it would require.

## Added to documentation?
- [ ] 👍 README.md

### DIFF
--- a/modules/DesktopAppearance/src/main/java/org/gephi/desktop/appearance/AppearanceUIController.java
+++ b/modules/DesktopAppearance/src/main/java/org/gephi/desktop/appearance/AppearanceUIController.java
@@ -49,6 +49,7 @@ import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
+import javax.swing.SwingUtilities;
 import org.gephi.appearance.api.AppearanceController;
 import org.gephi.appearance.api.AppearanceModel;
 import org.gephi.appearance.api.Function;
@@ -75,50 +76,13 @@ public class AppearanceUIController {
     protected final Map<String, Map<TransformerCategory, Set<TransformerUI>>> transformers;
     //Architecture
     protected final AppearanceController appearanceController;
-    private final CopyOnWriteArraySet<AppearanceUIModelListener> listeners;
+    private final CopyOnWriteArraySet<AppearanceUIModelListener> listeners = new CopyOnWriteArraySet<>();
     //Model
-    private AppearanceUIModel model;
+    private volatile AppearanceUIModel model;
 
     public AppearanceUIController() {
         final ProjectController pc = Lookup.getDefault().lookup(ProjectController.class);
         appearanceController = Lookup.getDefault().lookup(AppearanceController.class);
-        pc.addWorkspaceListener(new WorkspaceListener() {
-            @Override
-            public void initialize(Workspace workspace) {
-            }
-
-            @Override
-            public void select(Workspace workspace) {
-                AppearanceUIModel oldModel = model;
-                model = workspace.getLookup().lookup(AppearanceUIModel.class);
-                if (model == null) {
-                    AppearanceModel appearanceModel = appearanceController.getModel(workspace);
-                    model = new AppearanceUIModel(appearanceModel);
-                    workspace.add(model);
-                }
-                model.select();
-
-                firePropertyChangeEvent(AppearanceUIModelEvent.MODEL, oldModel, model);
-            }
-
-            @Override
-            public void unselect(Workspace workspace) {
-                if (model != null) {
-                    model.unselect();
-                }
-            }
-
-            @Override
-            public void close(Workspace workspace) {
-            }
-
-            @Override
-            public void disable() {
-                AppearanceUIModel oldModel = model;
-                model = null;
-                firePropertyChangeEvent(AppearanceUIModelEvent.MODEL, oldModel, model);
-            }
-        });
 
         if (pc.getCurrentWorkspace() != null) {
             model = pc.getCurrentWorkspace().getLookup().lookup(AppearanceUIModel.class);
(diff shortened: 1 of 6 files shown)


mbastian:
Reviewed. One correction to the description above: I originally wrote that the listeners on the `REFRESHING`/`REFRESHED` path already defer to the EDT themselves. That is true of `PreviewTopComponent` but not of `PreviewSettingsTopComponent`, whose `REFRESHING` branch calls `enableRefreshButton()`/`disableRefreshButton()` inline (`PreviewSettingsTopComponent.java:324-331` and `:346-353`, six unguarded `setEnabled()` calls each). `PreviewUIControllerImpl.refreshPreview()` fires those events from the `"Refresh Preview"` thread, so every Refresh click mutates them off the EDT.

It is a live bug but a different mechanism from workspace-event delivery, so it stays out of this diff and is now finding 11 on #3246.

